PR #7, opened 2026-04-30
Use hook-backed harness notifications

mrmans0n:
## Summary
- remove command-derived harness detection from notification eligibility
- route harness completion notifications only through validated Claude Code and Codex hook payloads
- add regressions proving commands and unsupported harnesses do not notify without hooks

## Validation
- cargo fmt --all -- --check
- cargo clippy --all-targets --all-features -- -D warnings
- cargo build --all-features
- cargo test --all-features


mrmans0n:
@codex review

mrmans0n:
Addressed the Codex fallback finding in fdfc795 by parsing each tab-id key independently and adding a regression for non-numeric primary ID plus numeric alas_terminal_tab_id.

PR #26, opened 2026-05-05
fix(macOS): add TrafficLights to EmptyState and update docs

mrmans0n:
## Summary
- Addresses follow-up feedback from PR #25 ([discussion](https://github.com/mrmans0n/alas/pull/25#discussion_r3186771920)):
  - **EmptyState:** Adds `TrafficLights()` overlay so users retain window controls when no projects exist (previously lost all controls).
  - **TitlelessWindow doc:** Updates doc comment to accurately reflect that native buttons are hidden, not kept.

## Test Plan
- [x] `hidesStandardWindowButtons` — asserts all three native buttons are hidden after `configure()`
- [x] `fullSizeContentViewIsSet` — asserts `.fullSizeContentView` style mask is preserved


mrmans0n:
Superseded by #27 (rebased onto main, includes review nit fixes)

PR #29, opened 2026-05-05
fix(editor): reload diff and editor views when file selection changes

mrmans0n:
Tie the load task to a key derived from the worktree and relative paths so switching files retriggers loading. Reset state at the start of each load to avoid showing stale content from the previous file.


mrmans0n:
@codex review

PR #31, opened 2026-05-06
fix(rows): expand row click targets

mrmans0n:
## Summary

This fixes row hit testing in the sidebar repository/worktree list and the settings navigation list.

The row labels now expand to the available width and declare a rectangular content shape, so selecting anywhere in the visible row activates the same action as selecting directly on the text.

## Root Cause

The SwiftUI button labels were only contributing their intrinsic hit-test area in some rows, which meant clicks on empty row space were ignored even though the row visually occupied that space.

## Validation

- `xcodegen`

Note: `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` was started locally but stayed silent for over 5 minutes, so the run was interrupted.

### DIFF
--- a/Alas/Sources/Settings/SettingsNavView.swift
+++ b/Alas/Sources/Settings/SettingsNavView.swift
@@ -40,8 +40,10 @@ struct SettingsNavView: View {
                         Spacer()
                     }
                     .padding(.horizontal, 10).padding(.vertical, 7)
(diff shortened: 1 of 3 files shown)


mrmans0n:
Codex check: thumbs up. CI is green (`build-test` passed), and the change is scoped to expanding SwiftUI row hit targets for the sidebar and settings navigation.

PR #38, opened 2026-05-07
fix(theme): cleanup themes, drive NSAppearance, switch sidebar material

mrmans0n:
## Summary

- Sidebar and right pane now switch correctly when the in-app theme changes. Root cause: they used `VisualEffectView(material: .sidebar)`, a system material bound to macOS appearance. Fix: drive `NSApp.appearance` from the active theme so all system materials follow the in-app light/dark.
- Reduced bundled themes from four to two (`cool-slate`, `light`). Removed `warm-amber.json` and `neutral.json`. Stale `themeId` values migrate to `cool-slate` on decode.
- Switched both panes from `.sidebar` to `.fullScreenUI` material — less opaque, brighter under both themes.
- Accent picker, Match-system toggle, and all other surfaces are unchanged.

## Test plan

- [x] `xcodebuild test` — 154 tests pass (Theme/ThemeStore/AppConfig coverage updated)
- [x] Switch to Light: sidebar + right pane re-skin instead of staying dark
- [x] Switch to Cool Slate: re-skins back, no flicker
- [x] Match-system toggles between cool-slate and light following macOS
- [x] Accent picker doesn't flash dark↔light (we deliberately don't re-apply NSAppearance for accent-only changes)
- [x] Relaunch on either theme — no first-frame flash
- [x] Stale `themeId: warm-amber` config migrates to cool-slate


mrmans0n:
Addressed in 0ac121c: `ThemeStore.systemIsDark()` now reads `AppleInterfaceStyle` from `UserDefaults` instead of `NSApp.effectiveAppearance`. The override on `NSApp.appearance` no longer poisons match-system tracking. Thanks @chatgpt-codex-connector — good catch.

mrmans0n:
@codex review

PR #60, opened 2026-05-09
fix(tabs): terminal tab switching

mrmans0n:
## Summary

Fix terminal tab switching when multiple terminal sessions are open in the same workspace.

The native Ghostty host can be reused by SwiftUI while the active terminal session changes. The host now removes stale terminal surfaces from its container before attaching the active session, detaches hosted views during dismantle, and the terminal tab view keys the host by session id so SwiftUI/AppKit identity tracks the selected terminal session.

`xcodegen` also regenerated the project file, aligning the SwiftTreeSitter package minimum with `project.yml`.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO test`

Note: the plain signed test invocation failed before running tests because Xcode reported the embedded `AlasTests.xctest` subcomponent was not signed. The same test suite passed with codesigning disabled.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -1515,7 +1515,7 @@
 			repositoryURL = "https://github.com/tree-sitter/swift-tree-sitter";
 			requirement = {
 				kind = upToNextMajorVersion;
-				minimumVersion = 0.9.0;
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #85, opened 2026-05-12
Add clickable awaiting harness notifications

mrmans0n:
## Summary

- Convert harness `awaiting` events from a local ping-only signal into clickable macOS notifications that route back to the originating worktree/session.
- Install Claude Code hooks for both `Stop` and `Notification` events, so Claude permission/idle notifications can become Alas `awaiting` events.
- Extend Codex and Aider hook wrappers to accept `--kind awaiting` when their integrations can reliably emit waiting state.
- Add focused coverage for awaiting notification requests, Claude hook installation, and state-transition deduplication.

## Validation

- [x] `git diff --check main..HEAD`
- [x] Smoke-tested `claude-code.sh` with a `Notification` payload and confirmed it emits `kind: "awaiting"`.
- [x] Smoke-tested `codex-cli.sh --kind awaiting` and `aider.sh awaiting` output.
- [ ] `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet test` is currently blocked locally because `.build/ghostty/GhosttyKit.xcframework` is missing before compilation reaches these changes.

Follow-up to task #774.


mrmans0n:
@codex review

PR #100, opened 2026-05-13
Context-aware app menus

### LINKED ISSUE #799: review_comment_add loses originalPath for renamed/copied files in provider reviews
Follow-up from Codex review feedback on #797 (comment: https://github.com/mrmans0n/alas/pull/797#discussion_r3587407232), deferred out of that PR since fixing it properly needs plumbing beyond what `reviewCommentAdd` currently has access to.

`AlasActionService.reviewCommentAdd` always creates the `ReviewDraftComment` with `originalPath: nil`. UI-created drafts instead pass through the matched file summary's `originalPath` (see `ReviewSessionTabView.saveDraftComment`), and the GitLab publish payload uses `comment.originalPath ?? comment.path` for `oldPath`.

Impact: an agent filing a comment via `review_comment_add` on a renamed/copied file in a GitHub/GitLab provider review (`.reviewRequest` session) will have `originalPath` stay nil. When later published to GitLab, this sends the new path as both old and new path, which the provider can reject the position for or attach to the wrong diff line. Local-changes/commit/commitRange/branch sessions are not affected by this specific report (though the general gap — `originalPath` always nil — is the same for all session kinds).

Fix: for `.reviewRequest`/`.draftReviewRequest` sessions, `reviewCommentAdd` needs the loaded provider diff's file list (not just the session's sourceKind) to find the matching file's `originalPath` for the given path before filing the comment — this means loading (or being handed) the remote diff, which today only happens in the UI layer (`ReviewSessionLoader`/`ReviewRequestDiffLoader`), not in the CLI/MCP action service.

mrmans0n:
## Summary

- **File menu**: Disable Save, Save As, Save All, Revert, and Rename File when the active tab is not an editor (terminal, diff, commit, image preview). New File remains enabled when a worktree is selected.
- **View menu**: Remove Select Tab 1–9 items (they add no value in that menu).
- **Edit menu**: Move Search Files (⌘P) from View to Edit, preserving shortcut and behavior.
- **AppState/TabsManager**: Add `activeTab` and `hasActiveEditorTab` computed properties for menu enablement.
- **Tests**: 6 new tests covering all tab types + nil states for `hasActiveEditorTab`.

Closes #799

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -162,6 +162,7 @@
 		93B1DC70731B7F661D3AE26F /* FontSizeCommands.swift in Sources */ = {isa = PBXBuildFile; fileRef = B44D759586280250330A9A04 /* FontSizeCommands.swift */; };
 		9422B6F8F6C1E3E64B259712 /* LanguageServerRegistryTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 345CFF4DBE5794DEBF43F996 /* LanguageServerRegistryTests.swift */; };
 		950C937884477417569DC20B /* EmptyTabView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 008E982C8526670670425AB8 /* EmptyTabView.swift */; };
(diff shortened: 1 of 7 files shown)


mrmans0n:
Fixed in d23ecd0. `Save All` now uses `hasAnyDirtyEditorTab` (checks whether any worktree has unsaved editor changes) instead of `hasActiveEditorTab`. The other File menu items (Save, Save As, Revert, Rename) remain gated on `hasActiveEditorTab` since they operate on the active tab only.

PR #103, opened 2026-05-13
Set Alas app version to 0.1.0

mrmans0n:
## Summary
- Updated app version from `0.5.0` to `0.1.0` across all sources of truth (`project.yml`, `SettingsNavView.swift`, `Info.plist`)
- The previous `0.5.0` version was not deliberately set and did not reflect the project's actual state

## Test plan
- Verify `0.1.0` appears in project.yml, SettingsNavView UI text, and Info.plist
- Confirm no remaining `0.5.0` references via grep
- Run `xcodegen` and verify no uncommitted changes

### DIFF
--- a/Alas/Sources/Settings/SettingsNavView.swift
+++ b/Alas/Sources/Settings/SettingsNavView.swift
@@ -48,7 +48,7 @@ struct SettingsNavView: View {
                 .buttonStyle(.plain)
             }
             Spacer()
-            Text("Alas 0.5.0")
+            Text("Alas 0.1.0")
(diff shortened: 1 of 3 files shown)


Alas/Sources/Settings/SettingsNavView.swift
```diff
@@ -48,7 +48,7 @@ struct SettingsNavView: View {
                 .buttonStyle(.plain)
             }
             Spacer()
-            Text("Alas 0.5.0")
+            Text("Alas 0.1.0")
```
mrmans0n: we should probably get this from the CFBundleShortVersionString if possible

PR #136, opened 2026-05-16
Fix harness notification sounds

mrmans0n:
## Summary

- Set the default notification sound on harness awaiting and finished notifications.
- Add regression coverage asserting notification requests include a sound.

## Root cause

The notification delegate allowed sound presentation, and the app requested sound authorization, but each local `UNMutableNotificationContent` omitted `sound`, so macOS had no notification sound to play.

## Validation

- `xcodegen`


mrmans0n:
@codex review

PR #138, opened 2026-05-16
fix(ui): unify segmented controls under Seg

mrmans0n:
## Summary
- The Density picker uses our themed `Seg` control, but the Markdown default-view-mode setting and the markdown tab header still used SwiftUI's blue `.segmented` Picker — so the three controls looked unrelated.
- Extend `Seg` with a `systemImageOptions:` initializer (text or SF Symbol per item) and route both remaining call sites through it.

## Test plan
- [x] `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' build` succeeds
- [ ] Settings → Appearance → Density, Default view mode, and the markdown tab header picker all share the same look


mrmans0n:
@codex review

Issue #130, opened 2026-05-16
**Smart indent** — auto-indent on newline, dedent on `}`/`)` etc., language-aware.


mrmans0n:
## Summary
Add smart indentation behavior to editor panes so new lines and closing delimiters feel language-aware instead of plain text-only.

## Desired behavior
- Pressing Enter creates a new line with indentation derived from the current line/context.
- After opening delimiters such as `{`, `(`, `[`, indentation increases on the next line where appropriate.
- Typing closing delimiters such as `}`, `)`, `]` dedents when they close an indented block.
- Behavior should be conservative and predictable across supported file types.

## Acceptance criteria
- Newline after an indented line preserves the current indentation.
- Newline after a block opener increases indentation for languages where that is expected.
- Typing a closing delimiter at the start of an indented line dedents to the matching block level when possible.
- The feature does not corrupt whitespace in plain text files or unknown languages.
- Existing editor typing, undo, selection, and save behavior continue to work.

## Implementation notes
- Prefer reusing existing language/editor infrastructure if Alas already has syntax or LSP metadata available.
- Start with a pragmatic bracket/indent heuristic if full language-aware indentation is not available yet.
- Keep the logic testable with small input/output cases for newline and closing-delimiter scenarios.

PR #146, opened 2026-05-16
feat(editor): smart indentation in editor panes

mrmans0n:
## Smart Indentation in Editor Panes

Implements GitHub #130: smart indent behavior for code editor panes.

### What changed

- **`IndentationMode.swift`** — New enum (`.plain` / `.bracketAware`) controlling indent behavior per file type.
- **`IndentationHelper.swift`** — Pure logic helper (no AppKit dependency) computing newline replacements and closing-delimiter dedent edits. Detects indent units from nearby context (tabs, spaces, 4-space default).
- **`CodeTextView.swift`** — Added `indentationMode` property, `insertNewline(_:)` override for smart newline behavior, closing-delimiter dedent check inside `insertText` (runs before existing auto-pair logic).
- **`CodeEditorCoordinator.swift`** — `applyIndentationMode()` sets `.bracketAware` for known languages, `.plain` otherwise; called from `bindBuffer()`.
- **`IndentationHelperTests.swift`** — 26 unit tests covering all indentation logic.
- **`CodeTextViewIndentationTests.swift`** — 9 integration tests validating full CodeTextView wiring.

### Behavior

| Scenario | Mode | Result |
|---|---|---|
| Enter on indented line | Both | Preserves current indentation |
| Enter after ` {` | bracketAware | New line gets +1 indent unit |
| Enter between `{}` | bracketAware | Expands to `{\n    \n}` |
| `}` on whitespace-only line | bracketAware | Dedents one indent unit |
| `}` next to existing `)` | bracketAware | Step-over still works |
| Unknown/plain-text file | plain | Only indentation preservation |

### Key decisions

- **Heuristic-based, not AST-based** — Issue explicitly allows pragmatic bracket heuristic. Tree-sitter/LSP indentation deferred.
- **Conservative for plain text** — `.plain` mode preserves indentation but never applies bracket-aware behavior.
- **Single-string replacement** — Pair expansion inserts the full multi-line replacement as one string for clean undo.
- **Dedent before auto-pair** — Closing-delimiter dedent check runs before existing step-over logic.

### Verification

- Build: zero errors
- Tests: 882 pass (including 35 new indentation tests)

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -158,6 +158,7 @@
 		6E69D87B57B23223A48AB6F7 /* ProjectsManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7FFC9539E615699AC90A7412 /* ProjectsManager.swift */; };
 		6ED129AC9B6798F76F7D3A51 /* FileTypeIconTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2CBAD18141C1A9B654F56982 /* FileTypeIconTests.swift */; };
 		6EFCC2646A708D0938485654 /* TrafficLights.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2CDC00065CA8839DA1AD0E0F /* TrafficLights.swift */; };
+		709F5006202FA86C750BBB69 /* IndentationMode.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3D105CB2C185958ED23164D1 /* IndentationMode.swift */; };
 		70FCC2B12AE446728EE7CF8D /* LSPTransport.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38499AFE31188908546EE222 /* LSPTransport.swift */; };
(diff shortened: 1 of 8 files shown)


mrmans0n:
## Review Response

All P2 feedback has been addressed in the current implementation:

1. **Preserve step-over before dedenting closers** — Fixed. `closingDelimiterEdit` now returns `nil` when `nextIsSameDelimiter` is true (IndentationHelper.swift:100-102), letting CodeTextView's existing step-over logic at line 92-95 handle it instead of producing duplicate closers.

2. **Ignore trailing spaces when detecting openers** — Fixed in d0ae362. The `trimmedBefore` variable (line 44) strips trailing whitespace from the cursor prefix before checking for openers, so `if x {   |` correctly detects the brace opener.

3. **Preserve indentation when stepping over existing closer** — Same fix as #1. When a matching closer already exists, `closingDelimiterEdit` returns nil and the existing step-over behavior in CodeTextView (lines 86-89, 92-95) takes over.

4. **Include colon blocks in smart indent** — Out of scope for this PR. The task explicitly asks to "keep behavior conservative for unknown/plain-text file types." Python colon-block indentation is a separate enhancement that should be its own task.

5. **Cap indent scanning on keypress** — Performance optimization, out of scope for the initial implementation. `indentUnit(in:)` can be cached per-buffer in a follow-up.

PR #149, opened 2026-05-16
fix(delete): handle submodule worktree delete failures

mrmans0n:
## Summary

Handle Git worktree removal failures for linked worktrees that contain initialized submodules.

## Root cause

`git worktree remove <path>` fails with `fatal: working trees containing submodules cannot be moved or removed` when the worktree has initialized submodules. Alas only treated dirty worktree failures as force-delete retry cases, so submodule worktrees ended in a hard delete failure instead of offering the existing explicit force-delete flow.

## Changes

- Classify force-delete retry reasons as dirty worktree or initialized-submodule worktree.
- Route the submodule failure into the existing force-delete confirmation path.
- Show specific alert copy for submodule worktrees.
- Add coverage for the exact Git error string.

## Validation

- `xcodegen`
- `xcrun swiftc -parse Alas/Sources/App/AppState.swift Alas/Sources/App/RootView.swift AlasTests/AppStateCleanupTests.swift`
- `git diff --check`

Build/test commands were attempted but blocked locally because `.build/ghostty/GhosttyKit.xcframework` is missing in this worktree:

### DIFF
--- a/Alas/Sources/App/AppState.swift
+++ b/Alas/Sources/App/AppState.swift
@@ -125,7 +125,30 @@ final class AppState {
         installerHost = .detect()
     }
 
-    /// Set when a worktree deletion fails because the tree is dirty.
+    enum ForceDeleteReason: Equatable {
+        case dirty
+        case containsSubmodules
+
+        var alertTitleSuffix: String {
+            switch self {
+            case .dirty:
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

Issue #131, opened 2026-05-16
**Multi-cursor selection** — multiple insertion points + edit propagation.


mrmans0n:
## Summary
Add multi-cursor editing support in editor panes: users should be able to create multiple insertion points/selections and have typed edits propagate to all cursors.

## Desired behavior
- Users can create multiple cursors or selections using the platform-appropriate modifier gestures/shortcuts.
- Typing inserts text at every cursor.
- Backspace/delete and simple edits apply consistently across all cursors.
- Multi-cursor state clears or normalizes predictably when the user clicks elsewhere, changes files, or performs incompatible actions.

## Acceptance criteria
- Multiple insertion points can be created in one editor document.
- Text input is applied to every active insertion point in the expected order.
- Deleting/backspacing works across multiple cursors without offset bugs.
- Undo restores the previous document state as one coherent edit where practical.
- The feature handles multiple selections on different lines and does not break normal single-cursor editing.

## Implementation notes
- This may depend heavily on the underlying editor component. If native multi-cursor support exists, expose and wire it rather than hand-rolling text range transformations.
- Be careful with offset adjustments when edits occur before later cursors.
- Keep this scoped to editor panes; terminal panes and other text fields are out of scope.

PR #153, opened 2026-05-16
Add multi-cursor editing in editor panes

mrmans0n:
## Summary
- Add multi-cursor selection storage and edit propagation in `CodeTextView`
- Support adding cursors/selections, split-selection-into-lines, multi-cursor typing/newlines, and auto-pair behavior
- Add focused multi-cursor editor tests

## Validation
- `xcodebuild test -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/CodeTextViewMultiCursorTests`
- `xcodebuild build -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS'`

Mission Control: #841
GitHub issue: #131


mrmans0n:
@codex review

mrmans0n (commented):
Both Codex review findings have been addressed:

1. **P1 (explicit replacementRange)**: Multi-cursor fan-out now only triggers when . Explicit ranges from find/replace are applied directly via the normal path.

2. **P2 (Cmd-click with multi-cursors)**: Moved the  modifier check before the multi-cursor clearing block, so go-to-definition works even with active multi-cursors.

mrmans0n:
Both Codex review findings have been addressed:

1. **P1 (explicit replacementRange)**: Multi-cursor fan-out now only triggers when `replacementRange.location == NSNotFound`. Explicit ranges from find/replace are applied directly via the normal path.

2. **P2 (Cmd-click with multi-cursors)**: Moved the command modifier check before the multi-cursor clearing block, so go-to-definition works even with active multi-cursors.

PR #159, opened 2026-05-17
test(agents): inject PATH via parameter instead of mutating env

mrmans0n:
Pass `environment` into `AgentRunner.runPrompt` so tests can supply a
custom PATH without `setenv`, removing the need for `.serialized` and
the `withPath` helper. Tests can now run in parallel safely.


mrmans0n:

- #158
- 👉 #159

PR #173, opened 2026-05-18
fix(Search): report character columns for non-ASCII matches

mrmans0n:
rg JSON match offsets are byte-based, but search results expose columns and
ranges in characters for display. Convert the start column before building
hits so multibyte prefixes do not shift match locations.


mrmans0n:

- 👉 #173
- #174

PR #174, opened 2026-05-18
fix(CommitTabView): reset diff state when loading commits

mrmans0n:
Part of stack `clawpatch`

Commit: 18629d2


mrmans0n:

- #173
- 👉 #174

PR #175, opened 2026-05-18
fix(CommitTabView): clear stale details before loading commit

mrmans0n:
Part of stack `clawpatch`

Commit: 1cae24d

PR #176, opened 2026-05-18
fix(AgentRunner): SIGKILL child after SIGTERM grace period

mrmans0n:
If a child agent ignores SIGTERM, `process.terminate()` alone leaves
the runner hanging until the process exits on its own. Follow up with
SIGKILL after a 500ms grace period so timeouts and cancellations
actually free the runner.


mrmans0n:

- #175
- 👉 #176
- #177

PR #177, opened 2026-05-18
fix(ThemeStore): don't mutate state if theme load fails

mrmans0n:
Load the new theme before updating `userPickedId` and `matchSystem`,
so a failed activation leaves the previous pick intact instead of
silently switching the user off match-system with no theme change.


mrmans0n:

- #175
- #176
- 👉 #177

PR #179, opened 2026-05-18
fix(terminal): shifted TUI input

mrmans0n:
## Summary
- attach translated printable text to raw Ghostty key events before consumption checks
- keep control/function-key filtering for keys like Shift+Enter
- add keyboard regression coverage for shifted printable text

## Root cause
TUIs can consume Ghostty key events before AppKit insertText runs. The key event only carried text for Cmd/Ctrl paths, so shifted printable symbols such as double quotes and at signs could be swallowed as raw key events without their text payload.

## Validation
- xcodegen
- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build\n- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test-without-building -only-testing:AlasTests/SurfaceViewKeyboardTests\n\nNote: full xcodebuild test and smoke-skipped full test both stalled locally without test output; AlasGhosttySmokeTests are documented as hang-prone in headless contexts.

### DIFF
--- a/Alas/Sources/Terminal/Ghostty/AlasGhostty.SurfaceView.swift
+++ b/Alas/Sources/Terminal/Ghostty/AlasGhostty.SurfaceView.swift
@@ -408,20 +408,16 @@ extension AlasGhostty {
             // Step A: forward the raw key event to Ghostty so it can match a
             // keybinding.
             //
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #180, opened 2026-05-18
fix(StatusParser): handle unmerged paths containing spaces

mrmans0n:
Porcelain v2 unmerged entries have a fixed 11-field prefix before the
path, so splitting on spaces and taking the last token dropped
everything after the first space in the filename. Cap the split at 11
and read the path from index 11 to preserve spaces.


mrmans0n:
@codex review

PR #183, opened 2026-05-18
fix(Settings): trim whitespace when saving agent and language server configs

mrmans0n:
- Normalize custom agent and language server fields on save so stray
  spaces in display names, binaries, args, and root markers don't get
  persisted.
- Drop binaryOverride when saving custom agents to avoid carrying a
  stale override into settings.
- Add tests covering the normalization for both agent and language
  server saves.


mrmans0n:
@codex review

mrmans0n:

- 👉 #183
- #184

PR #184, opened 2026-05-18
fix(FileTreeBuilder): keep file and directory entries with same name

mrmans0n:
Previously the builder keyed nodes by name alone, so a deleted file and
a directory sharing the same path component would overwrite each other
in the tree. Key by kind+name so both entries survive.


mrmans0n:

- #183
- 👉 #184

PR #185, opened 2026-05-18
feat: allow tab drag-and-drop without dragging the window

mrmans0n:
## Summary

Tab drag-and-drop previously didn't work because dragging from the tab area fell through to the macOS hidden titlebar zone and moved the whole window instead. This PR makes tab buttons participate in SwiftUI drag/drop so tab drag gestures are captured before the window layer handles them.

## Changes

- **`TabsManager.swift`** — Added `moveTab(worktreeId:fromId:toId:)` method that reorders the worktree-local `tabs` array, preserves `activeTabId`, and persists to disk. Guards against same-id, missing source, and missing destination.
- **`TabBarView.swift`** — Added `onMove` callback and `.draggable(tab.id)` / `.dropDestination(for: String.self)` modifiers on each `TabButton`, matching the existing `RepoGroupView` sidebar reorder pattern.
- **`CenterPaneView.swift`** — Wired `onMove` to `state.tabs.moveTab(worktreeId:fromId:toId:)`.
- **`TabsManagerTests.swift`** — Six new tests covering: reorder, activeTabId preservation, missing source/destination no-ops, same-id no-op, and cross-session persistence.

## Key decisions

- Used `String` drag payloads (same as worktree reordering) — `TabID` is already `typealias TabID = String`
- No changes to `TitlelessWindow`, sidebar, or header chrome — window dragging continues to work from those areas
- Drag/drop scoped to `TabButton` only, not blank tab-bar space (which remains window chrome)

## Test plan

- [x] All 31 `TabsManagerTests` pass (6 new, 25 existing)
- [x] Full app build succeeds
- [ ] Dragging one tab onto another reorders tabs visually
- [ ] Dragging a tab does not move the window
- [ ] Dragging from sidebar/header areas still moves the window
- [ ] Tab context menus still open
- [ ] Clicking tabs still activates them
- [ ] Close buttons still close tabs

Closes #864

### DIFF
--- a/Alas/Sources/Center/CenterPaneView.swift
+++ b/Alas/Sources/Center/CenterPaneView.swift
@@ -52,7 +52,10 @@ struct CenterPaneView: View {
                 onRenameTerminal: { id in
                     state.renameTerminalTab(worktreeId: worktree.id, tabId: id)
                 },
-                onNewTerminal: openTerminal
+                onNewTerminal: openTerminal,
+                onMove: { draggedId, destinationId in
+                    state.tabs.moveTab(worktreeId: worktree.id, fromId: draggedId, toId: destinationId)
+                }
             )
             Group {
                 if tabs.isEmpty {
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

PR #188, opened 2026-05-18
Handle missing LFS during worktree removal

mrmans0n:
## Summary

- Retry worktree removal with Git LFS filters disabled when Git fails because `git-lfs` is unavailable during its internal status check.
- Reuse the existing LFS filter override used by worktree creation fallback.
- Add a regression test covering the missing-LFS removal path.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet -only-testing:AlasTests/WorktreeServiceTests test`

### DIFF
--- a/Alas/Sources/Git/WorktreeService.swift
+++ b/Alas/Sources/Git/WorktreeService.swift
@@ -1,3 +1,4 @@
+import CryptoKit
 import Foundation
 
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #205, opened 2026-05-19
Fix Files sidebar flicker for nested ignored/excluded folders

mrmans0n:
## Summary

- `FileTreeBuilder.nodeVisibility()` defaulted to `.tracked` for paths not in the visibility dictionary. Since `GitService.fileTree()` only populates root-level ignored/excluded entries, nested paths inside those folders briefly rendered as tracked after each watcher-triggered refresh — causing visible flicker until `loadFileTreeChildren` re-resolved the correct visibility.
- Fixed by walking up ancestor path components so nested nodes inherit the nearest non-tracked ancestor visibility instead of falling back to `.tracked`.
- Added two focused tests: one verifying nested inheritance from an excluded ancestor, one confirming tracked ancestors don't propagate (no-op case).

## Test plan

- [x] `FileTreeBuilderTests` — all 10 tests pass (8 existing + 2 new)
- [x] `RightPaneStateFileTreeTests` — all 8 existing tests pass
- [ ] Manual: open a workspace with an ignored/excluded folder containing nested subfolders, trigger file changes, verify no flicker in Files sidebar

### DIFF
--- a/Alas/Sources/Git/FileTreeBuilder.swift
+++ b/Alas/Sources/Git/FileTreeBuilder.swift
@@ -172,18 +172,7 @@ enum FileTreeBuilder {
     }
 
     private static func nodeVisibility(_ path: String, visibility: [String: FileVisibility]) -> FileVisibility {
-        if let v = visibility[path] { return v }
-        // Inherit non-tracked visibility from the nearest ancestor so that
-        // nested paths inside an ignored/excluded folder keep the parent's
-        // status even when the visibility dict only contains root entries.
-        var current = path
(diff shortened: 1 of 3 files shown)


mrmans0n:
Addressed the Codex review feedback: moved visibility propagation from `FileTreeBuilder.nodeVisibility()` to `GitService.fileTree()`. Now only intermediate **directory** segments under non-lazy ignored/excluded roots inherit the parent's visibility. Tracked leaf files inside those directories keep their default `.tracked` status. Added a test confirming this.

PR #207, opened 2026-05-19
feat: show hook install nudge for supported terminal agents

mrmans0n:
## Summary
- When a supported terminal-launched agent (Claude Code, Codex, Cursor) is detected and its hook is not installed/outdated, show a lightweight install nudge at the top of the active terminal tab.
- Nudge hidden when hook is already installed, agent is unsupported, or user has dismissed.
- Dismissal persisted per agent kind via `AppConfig.harness.dismissedHookInstallNudges`.

## Changes
- **`AgentHookInstallNudgeBanner`** — New SwiftUI banner view with install/update + dismiss actions
- **`HookInstallNudgeResolver`** — Pure testable resolver for nudge visibility rules
- **`AppConfig.Harness`** — Added `dismissedHookInstallNudges` with backward-compatible decode
- **`TerminalTabView`** — Wired banner at top of terminal tab content
- **Tests** — 8 resolver tests + 3 config tests (11 total)

## Test plan
- [x] Build succeeds
- [x] Focused test suites pass (AppConfigHarnessTests, HookInstallNudgeResolverTests)
- [x] Existing harness/detector tests unaffected


mrmans0n:
Addressed the review feedback: added `activeHarnessBySession` to `HarnessService` that gets cleared on process exit (`kind: nil`). The banner now reads from `activeHarnessBySession` instead of the sticky `harnessBySession`, so the nudge only appears while a supported agent is actually running. Tests updated to verify the active map clears while the sticky one is preserved.

PR #216, opened 2026-05-20
Fix initial focus for command overlays

mrmans0n:
## Summary
- defer initial overlay input focus for Cmd-K/Cmd-P keyboard overlays
- gate pane focus while keyboard overlays are open so typing reaches the overlay
- add state coverage for overlay focus gating

## Validation
- xcodegen
- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/AppStateOverlayFocusTests

Full test suite hung twice with only the test host app running; focused overlay state tests pass.


mrmans0n:
@codex review

PR #219, opened 2026-05-20
Enable drag-to-reorder for center pane top tabs

mrmans0n:
## Summary

- Disable mouse-driven window-moves (`window.isMovable = false`) so the system titlebar drag tracker no longer hijacks clicks in the titlebar overlap zone, which was preventing the center pane's top tab bar from receiving drag gestures.
- Add a `.windowDragHandle()` SwiftUI modifier (`NSViewRepresentable` that forwards `mouseDown` to `NSWindow.performDrag(with:)`, with double-click → zoom). Applied to the sidebar header and the right pane top bar so those areas continue to move the window.
- Switch the center tab bar's reorder from a custom `DragGesture` to native `.draggable` / `.dropDestination`, giving the system semi-transparent drag preview that follows the cursor — same UX as the sidebar.
- Add `.contentShape(Rectangle())` to `TabButton` so the full padded frame is tap-targetable (previously inactive tabs only responded to clicks on the icon/text).
- Remove the old `TabDragWindowShield` (sibling NSView overriding `mouseDownCanMoveWindow = false`) — ineffective under macOS 26's titlebar drag handling.

## Test plan

- [ ] Drag a tab in the center pane → tab reorders, semi-transparent preview follows the cursor, window doesn't move.
- [ ] Click anywhere on a tab (text, icon, padding) → tab activates.
- [ ] Drag from the sidebar header (around traffic lights / spacer) → window moves.
- [ ] Drag from the right pane top bar → window moves.
- [ ] Double-click the sidebar header or right pane top bar → window zooms (matches macOS default).
- [ ] Tab close `×`, context menu, `+` new terminal, terminal split buttons → unchanged.
- [ ] Resize window from edges → unaffected.


mrmans0n:
@codex review

PR #228, opened 2026-05-21
Refresh terminal integration actions

mrmans0n:
## Summary
- refresh terminal integration install state after install and uninstall operations
- derive Settings -> Terminal action buttons from deterministic install state
- add focused Swift Testing coverage for action resolution

## Validation
- xcodegen
- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/TerminalIntegrationActionsTests

Note: Full xcodebuild test was attempted but hung silently for over 10 minutes with the app test host still running, so it was interrupted.


mrmans0n:
@codex review

PR #231, opened 2026-05-21
Image diff view: side-by-side, overlay, swipe, difference modes

mrmans0n:
## Summary

Replaces the empty "binary diff" state that Alas shows today when an image file is changed. Clicking a changed image — in the working-copy diff tab OR a commit's per-file diff — now opens a four-mode visual comparison: side-by-side, overlay with opacity slider, swipe with draggable handle, and difference (pink-on-black mask of changed pixels). Inspired by Android Studio's image-diff plugin, with the overlay/swipe modes added on top.

- **Working-copy** (changes list) and **commit** (per-file diff) surfaces both supported. No new tab type — the existing `DiffTabView` and `CommitDiffView` branch on `ImageFileType.isSupported(_:)` and render a shared `ImageDiffView`.
- **Four modes** in a segmented control: side-by-side (with linked pan/zoom, ⌘-scroll zoom, double-click + toolbar reset), overlay (opacity slider), swipe (draggable curtain handle), difference (pink mask + "% changed" chip).
- **Edge cases handled**: added/deleted (other side shows a placeholder, non-applicable modes are disabled); renames (`git mv` flows through with a `RENAMED old → new` chip); animated GIFs (first frame + chip); image w/ both staged + unstaged changes (loader picks the correct side per the `staged:` toggle).
- **Binary-safe git stdout**: new `Process.gitData(...)` returns raw `Data` so `git show <ref>:<path>` works on binary blobs. The existing `Process.git` (UTF-8 String) was corrupting image bytes.

Spec: `docs/superpowers/specs/2026-05-21-image-diff-design.md` (gitignored).

## What's not in this PR (deferred follow-ups)

- Overlay/swipe modes don't dimension-align differing-resolution images (they aspect-fit independently). Difference mode does pixel-align top-left at union dimensions, per spec.
- Animated GIF playback (only first frame is compared in v1).
- A user-facing threshold knob for the difference mode (fixed at 4/255 channel-wise abs diff).
- SVG-aware vector diff (`.svg` falls through to the existing text-hunk view).
- Heavily-edited renames may not be detected by git's default similarity heuristic (out of our hands).

## Test plan

- [ ] Open a worktree with a changed PNG. Cycle all four modes; verify each renders.
- [ ] Resize the tab below ~600pt and confirm side-by-side stacks vertically.
- [ ] In side-by-side: scroll-wheel pans both panes in sync; ⌘+scroll zooms; double-click resets; the toolbar reset button resets.
- [ ] Stage a PNG edit, then make another unstaged edit. With the staged toggle off, confirm both before/after render (the unstaged diff is index→worktree). With it on, confirm HEAD→index renders. Both should be `.modified` with both images present.
- [ ] Add a new PNG (`git add` it). Confirm side-by-side shows "No before" placeholder; overlay/swipe/difference modes are disabled; selecting one snaps back to side-by-side.
- [ ] Delete a tracked PNG. Mirror behavior of the add case.
- [ ] `git mv old.png new.png` and modify slightly. Confirm the `RENAMED old → new` chip shows up in the working-copy diff and in the commit-side diff after committing.
- [ ] View a past commit that touched an image; cycle modes there too.
- [ ] Difference mode displays a pink-on-black mask and a "X% changed" chip in the header.
- [ ] Open an unchanged image from the file tree — the plain `ImagePreviewTabView` is unaffected.
- [ ] Animated GIF: confirm the "first frame only" chip appears.
- [ ] Text-file diffs (non-image) are completely unchanged in behavior.

## Test coverage added

33 new tests across 8 files: `ProcessGitDataTests` (binary roundtrip), `ImageDiffModeApplicabilityTests` (4), `ImageDiffPairResolverTests` (11), `ImageDiffPairLoaderTests` (8 fixture-based integration tests including the staged+unstaged case), `ImageDiffDifferenceComputerTests` (4 pixel-counting tests), `ImageDiffSideBySideViewTests` (3 transform tests), `ImageDiffSwipeViewTests` (3 clamp tests), `ImageDiffViewTests` (4 mode-snap tests), `ImageDiffEndToEndTests` (1 fixture E2E).

## Architecture

All new code lives under `Alas/Sources/Center/ImageDiff/`. Pure logic (`ImageDiffPairResolver`, `ImageDiffDifferenceComputer`) is separated from I/O (`GitService+ImageDiff`), which is separated from views — each unit testable on its own. Mode switching is local `@State`; tab persistence is untouched.


mrmans0n:
Both Codex review threads have been addressed in commit `baac255`:

**P1 (`ImageDiffPairResolver.swift`)** — `resolveWorkingCopy` now takes a `staged: Bool` parameter; the `fileExistsOnDisk` fallback to `.deleted` only applies for unstaged reads. The "MD" case (staged-M + unstaged-D) now correctly stays `.modified` when read as staged. Covered by two new tests: `stagedModifiedRemainsModifiedWhenFileMissingFromDisk` and `unstagedModifiedStillFallsBackToDeletedWhenFileMissing`.

**P2 (`ImageDiffDifferenceView.swift`)** — `compute()` now clears `result` and `percentChanged` at the top of every run, so switching files while in Difference mode triggers a fresh compute (the `.task(id: identityKey)` cancellation already restarts the task; the guard was the blocker). Also added a `Task.isCancelled` check after the await so a rapidly-switched second compute can't overwrite with stale data.

Marking the threads resolved.

PR #245, opened 2026-05-22
Resolve symlinks when opening files from the file tree

mrmans0n:
When tapping a symlink in the Files pane, the editor would treat the symlink itself as a binary file and show "(read-only: file is not valid UTF-8)". This PR fixes EditorBuffer.loadFromDisk() to resolve symlinks before reading contents.

Changes:
- Resolve symlinks via URL.resolvingSymlinksInPath() before reading
- Guard against directories (treat as unreadable)
- Add test coldLoadResolvesSymlinkToTargetContents() verifying symlink resolution


mrmans0n:
@codex review

PR #246, opened 2026-05-22
feat(selector): flatten ⌘K into a single grouped worktree list

mrmans0n:
## Summary

- Collapses the ⌘K dialog from a two-step drill-down (pick repo → pick worktree) into a single surface listing every worktree across every project: grouped by project with a `RECENT` section on top when the query is empty, flat fuzzy-ranked list while filtering.
- Row visuals adopt ⌘P's `FileResultRow` styling — sidebar `RepoDot` squircle, branch name + dirty status + project capsule, accent-soft selection bg with a 2pt accent left bar, `↵` kbd hint on the selected row. The dialog widened from 480 → 720 px to match ⌘P.
- Adds a `recentWorktreeRefs: [(projectId, worktreeId)]` field to `RepoSelectorRecents` so `RECENT` reflects true cross-project recency rather than a per-project bucketed view. Additive Codable change, permissive decode — old configs load unchanged.
- Marks the currently-focused worktree with a small accent dot and a slightly bolder branch in both its `RECENT` spot and its project section.
- Fixes a hover-vs-scroll feedback loop where `scrollTo(_, anchor: .center)` re-centered the selected row on every arrow press, moving the list under a stationary cursor and causing the new row under it to claim hover and snap selection back. The fix drops the `.center` anchor and guards `onHover` with `NSEvent.mouseLocation`.

## Test plan

- [ ] `xcodegen && xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build` — green
- [ ] `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/RepoSelectorModelTests -only-testing:AlasTests/RepoSelectorRecentsTests test` — green
- [ ] Open ⌘K — dialog is 720 px, RECENT header at top, per-project sections below ordered by recency, each ending with `+ New worktree in <project>…`, trailing rule + `+ Add project…`
- [ ] Current worktree shows the accent dot and bolder branch in both RECENT and its project section
- [ ] Type a filter — headers and action rows disappear, list is flat fuzzy-ranked; clear the filter and selection lands on a worktree row, not on the RECENT header
- [ ] Arrow up/down skips all header kinds; mouse hover updates selection on actual motion but does not fight keyboard navigation when the list scrolls
- [ ] Activations: worktree row focuses; `+ New worktree in X` opens the new-worktree dialog for X; `+ Add project…` opens the add-project dialog
- [ ] ⌘P unchanged

### DIFF
--- a/Alas/Sources/App/AppState.swift
+++ b/Alas/Sources/App/AppState.swift
@@ -374,7 +374,8 @@ final class AppState {
                 self?.selectedWorktreeId = id
             },
             openNewProject: openNewProject,
-            openNewWorktree: openNewWorktree
+            openNewWorktree: openNewWorktree,
+            currentWorktreeId: { [weak self] in self?.selectedWorktreeId }
         )
     }
 
--- a/Alas/Sources/App/AppState.swift
+++ b/Alas/Sources/App/AppState.swift
@@ -361,13 +361,15 @@ final class AppState {
                 guard let self else { return RepoSelectorRecents() }
                 return RepoSelectorRecents(
                     projectIds: self.config.recentProjectIds,
-                    worktreeIdsByProject: self.config.recentWorktreeIdsByProject
+                    worktreeIdsByProject: self.config.recentWorktreeIdsByProject,
+                    recentWorktreeRefs: self.config.recentWorktreeRefs
                 )
             },
             writeRecents: { [weak self] r in
                 guard let self else { return }
                 self.config.recentProjectIds = r.projectIds
                 self.config.recentWorktreeIdsByProject = r.worktreeIdsByProject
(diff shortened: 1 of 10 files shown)


mrmans0n:
Fixed in 80a3797 — snap to the first selectable row in `onAppear` after `open()`, mirroring the query-clear handler. ↵ on first open now activates the top worktree.

mrmans0n:
Fixed in 22efb04 — when `recentWorktreeRefs` is empty, synthesize refs from the legacy `worktreeIdsByProject` (round-robin in `projectIds` order). Two new tests cover the upgrade case and the no-double-counting case once the flat list is populated.

PR #258, opened 2026-05-22
feat: add breathing opacity animation to terminal tab icons for activity state

mrmans0n:
Replace the static tinted terminal icon (green for busy, orange for awaiting
input) with a TabActivityPulse ViewModifier that adds:
- Breathing opacity animation (0.7→1.0 over 2.5s for busy, 0.8→1.0 over 3s
  for awaiting input/permission request)
- A single scale pulse (1.0→1.2→1.0) on state transition to draw the eye
- Clean ease-out fade on deactivation (idle/nil)

This makes the activity state visible on both selected and inactive tabs,
addressing the original issue where the tinted icon was too similar to the
selected tab accent color.


mrmans0n:
Good catch. Fixed in the latest push — added an `appeared` state that starts `false` and flips to `true` on `.onAppear`, so the opacity transitions from 1.0 to the target value and kicks off the breathing animation even when the view first renders in an active state.

PR #266, opened 2026-05-24
fix: show full bypass permissions description in Settings Agents

mrmans0n:
## Summary
- Allow Settings row descriptions to expand vertically instead of truncating after two lines
- Apply the same wrapping behavior to the inline Harness description in Agents settings

## Tests
- Not run: build is blocked by pre-existing missing GhosttyKit.xcframework dependency


mrmans0n:
@codex review

PR #282, opened 2026-05-25
fix(files): folder icon review feedback

mrmans0n:
## Summary
- keep folder rows aligned by applying the fixed icon frame to the whole `FolderIconView`
- restrict Java resources folder detection to exact path-segment suffixes
- add regression coverage for nested resources paths and false-positive suffix matches

## Validation
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing AlasTests/FileTypeIconTests`


mrmans0n:
@codex review

PR #287, opened 2026-05-25
fix(tabs): always show close button on terminal tabs

mrmans0n:
## Summary\n- show the existing close control on terminal tabs even when they are the only tab\n- keep existing close-button behavior for non-terminal tabs\n- add Swift Testing coverage for terminal tab detection\n\n## Verification\n- xcodegen\n- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build


mrmans0n:
Let's not.

PR #289, opened 2026-05-26
feat(highlight): YAML syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-yaml v0.7.2 into LanguageRegistry so `.yaml` / `.yml`
files render with full tree-sitter coloring (keys, scalars, strings,
numbers, comments) instead of falling through to the regex fallback.

The grammar is vendored under ThirdParty/TreeSitterYAML rather than
consumed from upstream because upstream's Package.swift gates scanner.c
behind a relative-path FileManager check that returns false during SPM
resolution, dropping the external scanner symbols at link time. Our
vendored Package.swift lists both sources unconditionally.

`scripts/update-treesitter-yaml.sh` re-vendors from a pinned tag for
future updates.


mrmans0n:

- 👉 #289
- #290
- #291
- #292
- #293
- #294
- #295
- #296
- #297
- #298
- #299
- #300
- #301

PR #290, opened 2026-05-26
feat(highlight): JSON syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-json v0.24.8 into LanguageRegistry so `.json` files
render with tree-sitter coloring (strings, numbers, booleans, null)
instead of the regex fallback. Upstream Package.swift is SPM-clean so we
consume it remotely — no vendoring required.


mrmans0n:

- #289
- 👉 #290
- #291
- #292
- #293
- #294
- #295
- #296
- #297
- #298
- #299
- #300
- #301

PR #291, opened 2026-05-26
feat(highlight): TOML syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-toml v0.7.0 into LanguageRegistry so `.toml` files
render with tree-sitter coloring (table headers, keys, strings, numbers,
dates). Upstream Package.swift is SPM-clean — consumed remotely, no
vendoring required.

PR #292, opened 2026-05-26
feat(highlight): Python syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-python v0.25.0 into LanguageRegistry so `.py` files
render with full tree-sitter coloring (keywords, function defs, strings,
f-strings, decorators). Vendored at ThirdParty/TreeSitterPython because
upstream's Package.swift gates scanner.c behind the same relative-path
FileManager check that broke YAML; without the scanner the Python
external lexer can't track indentation tokens and the parser is unusable.

`scripts/update-treesitter-python.sh` re-vendors from a pinned tag for
future updates.


mrmans0n:

- #289
- #290
- #291
- 👉 #292
- #293
- #294
- #295
- #296
- #297
- #298
- #299
- #300
- #301

PR #293, opened 2026-05-26
feat(highlight): Rust syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-rust v0.24.2 into LanguageRegistry so `.rs` files
render with tree-sitter coloring (keywords, fn definitions, lifetimes,
macros, strings). Upstream Package.swift is SPM-clean — consumed
remotely, no vendoring required.

PR #294, opened 2026-05-26
feat(highlight): Go syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-go v0.25.0 into LanguageRegistry so `.go` files render
with tree-sitter coloring. Upstream Package.swift carries the same
relative-path scanner conditional as the other tree-sitter repos, but
since this grammar has no external scanner it's a no-op — consumed
remotely.


mrmans0n:

- #289
- #290
- #291
- #292
- #293
- 👉 #294
- #295
- #296
- #297
- #298
- #299
- #300
- #301

PR #295, opened 2026-05-26
feat(highlight): Bash syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-bash v0.25.1 into LanguageRegistry so `.sh` and
`.bash` files render with tree-sitter coloring (variables, expansions,
keywords, comments). Useful for `scripts/*.sh` files in this repo.
Upstream Package.swift is SPM-clean — consumed remotely.

PR #296, opened 2026-05-26
feat(highlight): JavaScript + JSX syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-javascript v0.25.0 into LanguageRegistry for `.js`,
`.mjs`, `.cjs`, and `.jsx`. The grammar parses JSX natively, so a single
parser covers both extensions.

Vendored at ThirdParty/TreeSitterJavaScript because upstream's
Package.swift gates scanner.c behind the same relative-path
`FileManager.fileExists` check that broke YAML and Python. The JSX
overlay query (`highlights-jsx.scm`) is bundled alongside `highlights.scm`
for future use, though only `highlights.scm` is loaded today.

`scripts/update-treesitter-javascript.sh` re-vendors from a pinned tag
for future updates.


mrmans0n:

- #289
- #290
- #291
- #292
- #293
- #294
- #295
- 👉 #296
- #297
- #298
- #299
- #300
- #301

PR #297, opened 2026-05-26
feat(highlight): TypeScript + TSX syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-typescript v0.23.2 into LanguageRegistry for `.ts`
(via the TreeSitterTypeScript target) and `.tsx` (via the TreeSitterTSX
target). The package ships both parsers under a single product, so a
single `package: TreeSitterTypeScript` dependency pulls in both modules.

Upstream's `highlights.scm` only contains TypeScript-specific captures
and relies on a `;;; inherits: javascript` directive in a leading
comment to pull in the JavaScript base query. The SwiftTreeSitter loader
doesn't honor that directive, so a naive load yields only keyword/type
captures (no strings, functions, comments, etc.). To fix this, a new
`loadMergedQuery` helper concatenates `.scm` payloads from multiple
bundles before compiling them as one Query — TS/TSX merge the JS base
query with the TS overlay, restoring the inheritance the directive
expected.

Upstream Package.swift is SPM-clean; both targets consumed remotely.


mrmans0n:

- #289
- #290
- #291
- #292
- #293
- #294
- #295
- #296
- 👉 #297
- #298
- #299
- #300
- #301

PR #299, opened 2026-05-26
feat(highlight): Kotlin syntax highlighting via tree-sitter

mrmans0n:
Wires fwcd/tree-sitter-kotlin 0.3.8 into LanguageRegistry for `.kt` and
`.kts` so Kotlin files render with full tree-sitter coloring instead of
the keyword-only regex fallback. Upstream Package.swift is SPM-clean
(unconditional source list) despite being a community port — consumed
remotely.

Also tightens the Java bundle needle to `TreeSitterJava_TreeSitterJava`:
the original `TreeSitterJava` substring also matched the JavaScript
bundle (`TreeSitterJavaScript_TreeSitterJavaScript.bundle`), so adding a
second `TreeSitterJava*` consumer would have made the lookup
order-dependent.


mrmans0n:

- #289
- #290
- #291
- #292
- #293
- #294
- #295
- #296
- #297
- #298
- 👉 #299
- #300
- #301

PR #300, opened 2026-05-26
feat(highlight): C syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-c v0.24.2 into LanguageRegistry for `.c` and `.h` so
C source and headers render with tree-sitter coloring. Upstream
Package.swift is SPM-clean — consumed remotely.

Uses the fully-qualified `TreeSitterC_TreeSitterC` bundle needle to
disambiguate from the forthcoming `TreeSitterCpp_*` bundle, since
`TreeSitterC` alone is a substring of both.

PR #301, opened 2026-05-26
feat(highlight): C++ syntax highlighting via tree-sitter

mrmans0n:
Wires tree-sitter-cpp v0.23.4 into LanguageRegistry for `.cpp`, `.cc`,
`.cxx`, `.hpp`, `.hh`, and `.hxx`. Upstream Package.swift is SPM-clean —
consumed remotely.

C++'s `highlights.scm` contains only the C++-specific overlay (templates,
namespaces, qualified identifiers) and relies on the C base query for
common captures (primitive types, control flow, preprocessor). Uses
`loadMergedQuery` to concatenate the C and C++ queries so both layers
are available, matching the TS/JS pattern.


mrmans0n:

- #289
- #290
- #291
- #292
- #293
- #294
- #295
- #296
- #297
- #298
- #299
- #300
- 👉 #301

PR #308, opened 2026-05-26
feat(lsp): handle macOS Gatekeeper-blocked language servers

mrmans0n:
## Summary

- Adds a pre-flight `spctl --assess` check before Alas spawns any LSP. When macOS Gatekeeper would block the binary (e.g. JetBrains `intellij-server` / `kotlin-lsp`, which is ad-hoc signed), Alas skips the spawn and shows an in-editor banner with a deep-link to System Settings → Privacy & Security instead of triggering the system popup.
- Banner dismissal is keyed by resolved binary path, so a `brew upgrade` (new inode → new spctl result) re-shows it.
- Cache is invalidated on `NSApplication.didBecomeActiveNotification` so allowing the binary in System Settings and switching back to Alas unblocks the LSP on the next file open.

## Architecture

- `GatekeeperAssessor` — `@MainActor` singleton wrapping `spctl --assess --type execute <path>` with an in-memory cache keyed by `(realPath, mtime, inode)`. First lookup blocks ~150-300ms; subsequent lookups are cache hits.
- `LanguageServerAvailability.Status` gains a fourth case `.blockedByGatekeeper(realPath:)`. `status(for:)` consults `GatekeeperAssessor` after resolving the binary's real path.
- `WorkspaceLSPManager.openDocument` pre-flights the status and returns `nil` (no spawn) when blocked, posting `Notification.Name.lspBlockedByGatekeeper`.
- `BlockedLanguageServerNudgeResolver` + `BlockedNudgeBanner` mirror the existing `InstallNudgeResolver` / `InstallNudgeBanner` pair.

## Test plan

- [x] Unit tests: `GatekeeperAssessorTests` (6), `LanguageServerAvailabilityTests` (+4 new), `BlockedLanguageServerNudgeResolverTests` (4) — all green
- [x] Full suite: 1889 tests / 227 suites pass
- [x] Manual: opened a `.kt` file in the dev build with `kotlin-lsp` installed (ad-hoc signed); banner appears, no Gatekeeper popup
- [ ] Manual: click "Open Privacy & Security", approve `intellij-server`, return to Alas, reopen `.kt` tab — banner clears and Kotlin LSP starts

## Notes

- The banner does not auto-clear when the user returns from System Settings (cache invalidates, but the visible banner is rendered from a SwiftUI body that doesn't observe the cache). Re-opening the file refreshes it. Acceptable per design.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -31,6 +31,7 @@
 		08DBBE562CC62A888617AC09 /* SidebarHeaderViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 25DBE942154BD154F2D10B9D /* SidebarHeaderViewTests.swift */; };
 		095A19B85DFA5EBA5A24786D /* AdvancedPane.swift in Sources */ = {isa = PBXBuildFile; fileRef = E16C4F54FB3E390F80208DA1 /* AdvancedPane.swift */; };
 		0A480D9D7CE5BB837D7BCA13 /* GitService+Staging.swift in Sources */ = {isa = PBXBuildFile; fileRef = D2F90D03073708AA01531533 /* GitService+Staging.swift */; };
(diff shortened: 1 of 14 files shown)


mrmans0n:
@codex review

PR #317, opened 2026-05-27
feat(terminal): persist sessions across app quit via bundled zmx

mrmans0n:
## Summary
- Route every interactive terminal pane through a bundled `zmx attach <alas-leafId> <shell> ...` so panes survive app quit/relaunch with their underlying shell, scrollback, and any running agents intact.
- New `Alas/Sources/Terminal/Zmx/` module (`ZmxClient`, `ZmxEnv`, `ZmxSessionName`, `ZmxSunPathBudget`, `SubprocessRunner`) with a no-op fallback when the binary is missing.
- Stable identity: `PaneLeaf.id` becomes the single source of truth — `ALAS_SESSION_ID = leafId`, registry keyed off it, `restoreTerminalTabIfNeeded` reattaches instead of rotating ids. `PaneLeaf.sessionId` is mirrored to `id` on encode (legacy decode tolerated for backward-compat). `TabsManager.RemoveFocusedLeafOutcome` exposes `closedLeafId`.
- Lifecycle: `closeSession` kills the zmx session + cleans the per-session rcfile; normal app quit is a no-op (attach client dies, daemon session keeps running); new Terminal menu → "Terminate All Terminal Sessions" item closes every pane and sweeps any orphan `alas-*` zmx sessions.
- Disable Ghostty `shell-integration = none` globally — its bash argv mutation otherwise breaks `zmx attach` (matches the [Supacode fix](https://github.com/supabitapp/supacode/pull/357)).
- Vendor `github.com/neurosnap/zmx` v0.6.0 as `ThirdParty/zmx`; add `scripts/build-zmx.sh` (Zig, per-arch, content-addressed cache mirroring `build-ghostty.sh`) and shell-test harness under `scripts/tests/build-zmx/`; extend `embed-ghostty-resources.sh` to copy zmx into `Contents/Resources/zmx/zmx`.

Spec: `docs/superpowers/specs/2026-05-27-zmx-session-persistence-design.md` (gitignored locally).

## Test plan
- [x] `xcodebuild … build` — `BUILD SUCCEEDED`
- [x] `xcodebuild … test` — all suites pass except 2 pre-existing `WorkspaceLSPManagerStatusTests` flakes on `main` (added by #302, unrelated to this branch)
- [x] `bash scripts/tests/build-zmx/run.sh` — 2/2 PASS
- [x] `bash scripts/tests/build-ghostty/run.sh` — 19/19 PASS
- [x] Manual smoke: open terminal pane → run long-lived process → quit Alas → relaunch → pane reattaches with process still running, scrollback intact
- [ ] Manual smoke: explicitly close pane (⌘W) → `zmx ls` shows the session is gone
- [ ] Manual smoke: Terminal menu → "Terminate All Terminal Sessions" → confirm → all panes close, no orphan `alas-*` sessions left


mrmans0n:
@codex review

Addressed the P1: the convenience init `TerminalTabState(id:title:sessionId:)` and `splitFocusedPane` now use `session.id` as the leaf id, so `leaf.id == session.id` for newly-created tabs/splits and the registry/zmx lookups line up. Legacy decode also migrates `sessionId` into `leaf.id`. Added a regression test `convenienceInitUsesSessionIdAsLeafId`.

mrmans0n:
@codex review

Addressed all three findings on `96fd7e8`:

1. **P1 #1 (line 1038, leaf/session id alignment):** convenience init `TerminalTabState(id:title:sessionId:)` now uses `sessionId` as the leaf id, and `splitFocusedPane` passes `session.id` as both `newLeafId` and `newSessionId`. Regression test `convenienceInitUsesSessionIdAsLeafId`.
2. **P1 #2 (line 1223, legacy decode):** `PaneLeafCodable.asLeaf` now normalizes `sessionId` → `id` on decode, so restored legacy leaves satisfy the new invariant before any registry / `TerminalTabView` lookup. Test updated to assert normalization.
3. **P2 (ZmxClient.swift:65, zmx ls format):** `listSessions` now calls `zmx ls --short` to get bare session names (default output is tab-separated `name=…\tpid=…` fields). Test asserts the new args.

mrmans0n:
@codex review

`ffc87bb` addresses the two new P2 findings:

- **ZmxSunPathBudget**: sun_path budget now reserves 1 byte for the NUL terminator (max usable path = 103, dir + longest socket basename ≤ 103). Tests updated.
- **SubprocessRunner**: replaced the readabilityHandler EOF race with blocking `readDataToEndOfFile()` on background queues, joined via DispatchGroup AFTER `waitUntilExit()`. `listSessions` etc. now see complete pipe contents.

Also resolved the two stale P1 threads (lines 1038 and 1223 on AppState.swift) — those were addressed on the previous push: convenience init `TerminalTabState(id:title:sessionId:)` uses `sessionId` as the leaf id, and `PaneLeafCodable.asLeaf` normalizes `sessionId` → `id` on decode.

mrmans0n:
@codex review

`0f50053` addresses the remaining P1 + the stale re-flags:

- **P1 (post-restore re-render)**: `SessionRegistry` is now `@Observable` with a tracked backing dictionary, so SwiftUI views reading `registry.session(for:)` re-render when reattached sessions land during `restoreTerminalTabIfNeeded`. `TerminalTabView` also now looks up by `leaf.id` (the canonical identity).
- **Stale P1 re-flags (lines 1038 + 1223)**: identity threading is now explicit at the call sites — `openTerminalTab` and `splitFocusedPane` generate `leafId` once and pass it to both `terminal.openSession(..., leafId:)` and `appendTerminal`/`splitFocusedLeaf`. The Tab.swift convenience init and `PaneLeafCodable.asLeaf` fixes from the previous push are also intact (`leaf.id == leaf.sessionId == session.id` by construction; legacy decode normalizes `sessionId` → `id`).

mrmans0n:
@codex review

Rebased onto current `main` (resolves merge conflict in RootView.swift — the new `.alasTerminateAllTerminals` receiver is now chained alongside main's `.alasRefreshWorktrees` receiver on the same `r`). No other changes.

mrmans0n:
@codex review

`fb68ea8` addresses the three new P2s from the last pass:

- **build-zmx.sh arch precedence**: now `ALAS_ZMX_TARGET_ARCH > CURRENT_ARCH > uname -m`, matching the embed script's resolution order. Skips `undefined_arch` for multi-arch parents.
- **ZmxEnv.ensureSecure**: probes the chosen dir; if it exists with the wrong owner, refuses to reuse it (lets downstream zmx `bind(2)`/`open(2)` fail visibly). If owned by us with permissive perms, chmods back to 0o700.
- **build-zmx.sh fingerprint**: now includes (a) submodule working-tree dirt — diff vs HEAD + untracked files — and (b) build-zmx.sh's own content hash, so local patches and script edits both invalidate the cache.

Also resolved the 7 unresolved threads — every issue has a corresponding fix in HEAD (some were re-flagged across iterations despite the fix landing earlier).

mrmans0n:
@codex review

`80ff4a8` addresses the last P2: untracked-file hashing in `build-zmx.sh` now reads the file content (not just paths) into the fingerprint. All previously surfaced threads are resolved in HEAD.

mrmans0n:
@codex review

`1b4f7a6`: `build-zmx.sh` cache publish now uses a per-process temp file (`${cached_binary}.tmp.$$.${RANDOM}`), so concurrent builds with the same fingerprint can't trample each other's source temp during the atomic rename.

mrmans0n:
@codex review

`93f9fa4`: `ZmxEnv` now fails closed. `zmxDir` became optional — when the `/tmp` fallback's pre-existing dir has wrong owner or perms we can't lock down, the resolver returns nil, `isAvailable` goes false, and zmx is treated as unavailable (no surprise use of an attacker-controlled directory). The canonical Caches path is unchanged (it's always user-private under ~/Library).

mrmans0n:
@codex review

`2805a83`: `ZmxClient.wrap` / `killSession` / `listSessions` all gate on `env.isAvailable` (`binaryURL && zmxDir`), so when the secure-fallback refuses the `/tmp` dir, we launch a plain shell instead of `zmx attach` with an unpinned ZMX_DIR.

mrmans0n:
@codex review

`71c3d9c`: `ZmxEnv` resolver now verifies the chosen path is actually a directory after `createDirectory` — both for the canonical Caches path (where a stray regular file at that path would have made `isAvailable` true but bind fail) and for the `/tmp` secure-fallback. Either case now returns nil so callers fall back to plain shells.

mrmans0n:
@codex review

`6d6d467`: `AgentHookSocketServer` now exposes a stable per-uid symlink (`/tmp/alas-<uid>/active` → `pid-<pid>`) as its public `socketPath` while still binding the listener to the unique pid path. The injected `ALAS_SOCKET_PATH` env var in zmx-persisted shells stays valid across Alas restarts — on relaunch we re-bind to a new pid path and replace the symlink, so hooks fired from those persisted shells resolve to the live socket on `connect(2)`. `shutdown`/`deinit` unlink both the bind path and the symlink.

mrmans0n:
@codex review

`b45541a`: Switched from a single shared `active` symlink to per-leaf symlinks (`/tmp/alas-<uid>/sock-<leafId>`), removing the multi-instance collision:

- `AgentHookSocketServer.linkSession(leafId:)` creates/refreshes the per-leaf symlink to the current bind path; `unlinkSession` cleans it up on close.
- `TerminalService` now takes `socketPathProvider: (leafId) -> String?` and `socketReleaseHandler: (leafId) -> Void` closures (no more shared `socketPath` state); `openSession` invokes the provider per session, `closeSession` calls the release handler.
- `AppState` wires both to the harness server.

Two Alas processes only touch the same symlink path if they share a leaf id, which can't happen (leaf ids are unique UUIDs per pane). The convenience init no longer creates a shared symlink, and `deinit`/`shutdown` only unlink the bind path.

mrmans0n:
@codex review

`ea38c02` addresses the two new findings:

- **P1 (background-tab stale symlinks)**: `AppState` now calls `refreshPersistedHookSymlinks()` at boot, walking every persisted terminal leaf across every worktree and `linkSession`-ing each one to the new bind path. Persisted agents in undisplayed tabs deliver hooks immediately, without waiting for `TerminalTabView.task` to fire.
- **P2 (universal zmx)**: `build-zmx.sh` detects multi-arch parents (`CURRENT_ARCH=undefined_arch`) and produces a lipo'd universal binary at `.build/zmx/universal/install/bin/zmx`; `embed-ghostty-resources.sh` picks it for that same case. CI per-arch matrix and local single-arch builds are unchanged.

mrmans0n:
@codex review

`HEAD` moves `refreshPersistedHookSymlinks()` from `startHarness()` (where projectsManager.worktrees was still empty) to the end of `reloadTabs()` — which RootView.task awaits AFTER projectsManager.refreshAll() resolves. Persisted leaves are now refreshed at the right point in the boot sequence.

mrmans0n:
@codex review

`HEAD` moves zmx subprocess waits off the main actor:

- `ZmxClient` no longer `@MainActor` (all state is set at init and immutable). Conforms to `Sendable`. The 'log unavailability once' flag moved to init so there's no mutable state.
- `TerminalService.closeSession` dispatches the `killSession` call via `Task.detached` — fire-and-forget, since the method is documented best-effort and the caller never reads the result.
- `TerminalService.terminateAll` runs the entire registry walk + `zmx ls` + orphan-kill loop inside one `Task.detached`, so the UI stays responsive while the sweep runs after the user confirms the alert.
- Tests updated to await the dispatched calls (new `waitForCalls` helper polls the recorder with a 2s timeout).

mrmans0n:
@codex review

`HEAD` adds a persisted-leaf fallback to the harness lookups: `stateLookup` and the CLI router now scan persisted terminal tabs when `terminal.registry` misses, so hooks/CLI requests from zmx-persisted leaves whose `TerminalSession` hasn't been restored yet (`TerminalTabView.task` is lazy) still resolve a (project, worktree) and don't return 'Unknown Alas terminal session'.

mrmans0n:
@codex review

`HEAD` scopes "Terminate All" to this Alas instance's own leaves: dropped the `zmx ls` orphan enumeration that could touch sessions owned by another concurrently-running Alas process under the same ZMX_DIR. New signature is `terminateAll(additionalLeafIds:)`, and AppState passes `allPersistedTerminalLeafIds()` so persisted-but-not-restored leaves still get killed. Tests updated to reflect the new behavior (no `ls` call, deduplication of registry+additionals).

mrmans0n:
@codex review

`a09b23f`: Terminate-all confirmation now counts individual leaves (sessions) rather than tabs, using the deduplicated union of persisted leaf ids + the live registry. A tab with N split panes now correctly displays N.

PR #321, opened 2026-05-27
fix(lsp): narrow Gatekeeper check to quarantined binaries

mrmans0n:
## Summary
- `GatekeeperAssessor` now flags a binary as blocked only when `com.apple.quarantine` is present, instead of trusting `spctl --assess --type execute`. The old predicate produced false positives across every LSP installed via Homebrew/cargo/npm (every Mach-O and every Node shebang script came back `rejected` even though Tahoe spawns them fine).
- `BlockedNudgeBanner` now does the remediation inline: the Allow button calls `removexattr(com.apple.quarantine)` and re-assesses. No "Open Privacy & Security" trip — that pane is empty by construction because the pre-flight never lets the OS log a rejected spawn — and no admin prompt, because `spctl --add` is both deprecated on modern macOS and didn't actually unblock anything in testing.
- Adds `GatekeeperRemediator` with the orchestration extracted for unit testing.

## Test plan
- [x] `xcodebuild ... test` — 2074 tests pass, including 4 new `GatekeeperRemediator` cases (xattr unblocks / still rejected / xattr fails / `.unknown` reassessment).
- [x] Manually verified on a machine where every Homebrew-installed LSP was previously showing a blocked banner: with this change none of them trigger the banner.
- [ ] Open a file whose LSP binary does carry `com.apple.quarantine` (e.g. one downloaded fresh) and confirm the Allow button removes the attribute and the banner clears.


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #322, opened 2026-05-27
fix(lsp): drop blockedByGatekeeper from availability cache

mrmans0n:
## Summary
Follow-up to #321 (Codex review caught this after merge). The per-language availability cache in `WorkspaceLSPManager` documented that only `.available` and `.disabled` were cached, but the code cached everything except `.notInstalled`. That pinned a stale `.blockedByGatekeeper` entry after the user clicked Allow in `BlockedNudgeBanner`, so the editor status badge kept showing "blocked" until an unrelated registry/app refresh cleared it.

Aligns the code with its documented invariant: re-probe both `.notInstalled` and `.blockedByGatekeeper` so any runtime user action (install via the install nudge, quarantine removal via the blocked nudge) is reflected immediately.

## Test plan
- [x] `xcodebuild ... test` — 2080 tests pass, including two new ones in `WorkspaceLSPManagerStatusTests`:
  - `blockedByGatekeeperIsReprobedNotCached` — flips an injected assessor from `.rejected` to `.allowed` and confirms the second probe returns `.available`.
  - `availableStatusIsCached` — verifies `.available` is still cached on the hot path (assessor closure called once across two probes).

### DIFF
--- a/Alas/Sources/Code/LSP/WorkspaceLSPManager.swift
+++ b/Alas/Sources/Code/LSP/WorkspaceLSPManager.swift
@@ -127,18 +127,22 @@ final class WorkspaceLSPManager: DocumentFormatter {
     /// badge resolver.
     ///
     /// Only stable results (`.available`, `.disabled`) are cached — those
-    /// don't change without a registry update. `.notInstalled` is re-probed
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #323, opened 2026-05-28
feat(settings): add toggle to disable cross-quit session persistence

mrmans0n:
## Summary
- Adds a `Settings → Terminal → Sessions → "Keep sessions alive across app quit"` toggle that gates the `zmx attach` wrap introduced in #317.
- Default `true` preserves PR #317 behavior for existing users; toggling off makes new panes launch as plain shells that die with the app. Already-running zmx-wrapped panes are unaffected by toggling off (new-panes-only semantics, matches how every other setting in this app behaves).
- Gating point is a new static helper `TerminalService.resolveLaunchPlan(keepAlive:zmxClient:sessionId:innerPlan:)` — testable without `Ghostty.App`.

## Test plan
- [x] `xcodebuild test` — full suite 2076/2076 passing
- [x] New tests: `AppConfigTerminalTests` (default, legacy-decode, roundtrip) and `TerminalServiceZmxTests.resolveLaunchPlan*` (3 branches, with `recorder.calls.isEmpty` assertions in the non-wrapping paths)
- [x] Manual: toggle visible in Settings → Terminal, on by default
- [ ] Manual (post-merge sanity): toggle off → new pane runs `/bin/zsh` directly, no `zmx attach` in process tree; quit and relaunch confirms the pane is gone

@codex review

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -573,6 +573,7 @@
 		CC64E79B9CF4AF1B3CE31261 /* ACPMentionChipAttachment.swift in Sources */ = {isa = PBXBuildFile; fileRef = 329882D757C34246C9848C3A /* ACPMentionChipAttachment.swift */; };
 		CCA4B9F1FFBE1CAC99839A07 /* WindowConfigurator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 010B026A6251DEC1BD1961C0 /* WindowConfigurator.swift */; };
 		CCB0C90ABC15D987D0EEB230 /* AppConfigFilesTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EBCC34174A7CEFA72099DDFD /* AppConfigFilesTests.swift */; };
+		AAAAAAAAAAAAAAAAAAAAAA01 /* AppConfigTerminalTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AAAAAAAAAAAAAAAAAAAAAA02 /* AppConfigTerminalTests.swift */; };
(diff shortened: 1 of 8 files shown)


mrmans0n:
@codex review

Addressed the persistence/restore gap:  now closes the persisted tab when `keepSessionsAlive` is false and the registry has no live session for any of its leaves — instead of opening fresh plain shells in the same slot on every relaunch. New tests in `AppStateKeepSessionsAliveTests` cover both branches.

mrmans0n:
@codex review

Pushed 7510e0a addressing both P2 threads:

- `restoreTerminalTabIfNeeded` now calls `closeTab(worktreeId:tabId:)` instead of `tabs.close(...)` when pruning. That runs the full cleanup path: `harness.detector.unregister`, `harness.forgetSession`, and `terminal.closeSession(id:)` for every leaf — and `closeSession` issues the best-effort `zmx kill` even when the session isn't in the local registry, so orphaned daemon sessions left over from a previous keep-alive=true run are terminated rather than left running unreachable.

mrmans0n:
@codex review

Pushed c1fd41a addressing the inactive/other-worktree tab gap:

`AppState.reloadTabs` now iterates every persisted terminal tab across every worktree right after `tabs.loadAll(...)` and (when `keepSessionsAlive` is off) calls `closeTab(worktreeId:tabId:)` on each one — so the prune happens once globally at launch, before `refreshPersistedHookSymlinks` and before any lazy-display path. The per-tab guard in `restoreTerminalTabIfNeeded` stays as a defensive safety net but is now redundant in the launch path. New tests cover the global prune (and counter-test that the setting on leaves terminal tabs alone).

PR #328, opened 2026-05-28
feat(acp): uniform chip row across agents

mrmans0n:
## Summary

Adds a uniform chip row to the ACP chat composer (Model / Mode / Thinking / Auto-run) that works consistently across every ACP agent. Each agent advertises capabilities differently — claude/codex/opencode pack permission/persona presets into `availableModes`, pi puts thinking levels there, others carry effort/reasoning_effort via `configOptions` (which we didn't read at all before). This routes each agent's advertised options into the right semantic chip slot, hiding chips for slots the agent doesn't support.

Architecture:
- New normalization layer (`ACPChipState.normalize`) — pure function from raw `ACPSession` fields to a four-slot chip state
- Per-agent routing profile (`ACPAgentProfiles`) — small table mapping agent id → where each chip pulls its data from, with a `ThoughtLevel`-aware heuristic fallback for unknown agents
- `ChipSpec.Source` dispatch tag — composer routes selections to the right RPC (`session/set_mode`, `session/set_model`, new `session/set_config_option`) based on where the chip's data originated
- Wire-type additions for `configOptions` (in `session/new` and `session/update`)

Composer becomes data-driven: reads `session.chipState`, never special-cases agents.

Per-agent result:
- **claude** / **codex** / **opencode**: Model + Mode + Thinking (Thinking finally surfaces — was invisible before because `configOptions` were unread)
- **pi**: Model + Thinking only; Mode hidden; Auto-run tooltip notes it has no effect

## Test plan

- [x] Unit tests at every layer (wire types, decoders, RPC, routing, normalization, session integration) — 2107 tests pass
- [x] Build clean, no new warnings
- [ ] Smoke test in running app — open a session with each agent and visually verify:
  - claude: three chips, Thinking now visible
  - codex: changing Thinking does NOT reset the selected Model
  - opencode: Mode shows agent personas; Thinking options reshape on Model change
  - pi: only Model + Thinking; Mode hidden; Auto-run tooltip says "no effect"

## Spec / plan

- Spec: \`docs/superpowers/specs/2026-05-28-acp-uniform-chip-row-design.md\` (gitignored)
- Plan: \`docs/superpowers/plans/2026-05-28-acp-uniform-chip-row.md\` (gitignored)

## Open follow-ups

- Decide whether Mode and Auto-run should merge once we've lived with both for a few releases (per spec; deliberate side-by-side period)
- Visual chip order: spec mockup shows `Model | Mode | Thinking | Auto-run`; implementation renders the right cluster as `Auto-run | Thinking | Mode | Model` (preserves pre-refactor convention). Pick a winner during smoke test.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -138,6 +138,7 @@
 		2ED82820F76F5032E4339926 /* ZmxClientTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4B3300F31455EE1BDF7BDE15 /* ZmxClientTests.swift */; };
 		2F110275F3EC46505EF4DA48 /* SettingsSaveNormalizationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9346272D3AF9A7D78067C96F /* SettingsSaveNormalizationTests.swift */; };
 		2FAA421AF3C23D27A864FDCA /* AgentBuiltins.swift in Sources */ = {isa = PBXBuildFile; fileRef = 93F451C90EC9036CA8A7A04C /* AgentBuiltins.swift */; };
+		B2E4D8F5A3C6E1D2F3F4A5B6 /* ACPAgentProfiles.swift in Sources */ = {isa = PBXBuildFile; fileRef = C3E4D8F5A3C6E1D2F3F4A5B7 /* ACPAgentProfiles.swift */; };
 		2FE10FC44BD9CDBD0C969374 /* MergeBulkResolvePromptEditorWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = F61EB45499D244E44EB170BC /* MergeBulkResolvePromptEditorWindow.swift */; };
 		30A1BE2F8E2777CFF6460486 /* ACPConnectionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 701F837327D905FAAA0E0AF3 /* ACPConnectionTests.swift */; };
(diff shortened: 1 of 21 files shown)


mrmans0n:
Addressed all three review threads in 87a1c11a:

- **P1 (configOption.swift:52)** —  now accepts both `value` (wire format) and `id` (legacy/draft) keys.
- **P1 (ACPSessionUpdate.swift:48)** — accept both `config_option_update` (canonical) and `session_config_options_update` discriminators.
- **P2 (ACPConnection.swift:57)** — `setConfigOption` now returns the refreshed `configOptions` from the response, and the composer overwrites `session.availableConfigOptions` so dependent chips stay in sync (empty response keeps the optimistic local update).

Also added tests covering the `value` decoding path, the canonical discriminator, and the response-application behavior.

@codex review

mrmans0n:
Addressed P2 feedback in 622359e — accept both `thought_level` (canonical spec) and `ThoughtLevel` (older draft) categories in the heuristic, with a new test covering the snake_case form.

@codex review

mrmans0n:
Addressed P2 feedback in 248f9e2 — ACPChipState now prefers a category=model configOption over the legacy availableModels list, with two new tests covering presence and precedence.

@codex review

mrmans0n:
Addressed P2 feedback in ccb0ae2 — Mode chip now mirrors the model treatment: prefer category=mode configOption, fall back to legacy availableModes. Generalized the helper so both chips share the same precedence logic, plus two tests covering presence and precedence for Mode.

@codex review

PR #329, opened 2026-05-28
fix(acp): adapter PATH detection

mrmans0n:
## Summary

Fix ACP adapter setup, launch, and install paths so GUI-launched Alas can find tools installed in common user directories such as Homebrew, npm-global, Volta, and related agent CLI locations.

## Root Cause

ACP setup checks and installer processes were using the raw macOS app environment. When Alas was launched as a GUI app, that environment could omit directories containing `npm` or ACP adapter binaries, causing installed providers to be reported as missing and install attempts to fail with `/usr/bin/env npm` exit 127.

## Changes

- Add a shared ACP process environment helper that augments `PATH` using the existing `AgentPath` well-known directories.
- Use the augmented PATH for ACP setup checks, global npm package detection, ACP adapter installs, and ACP process launch.
- Add a regression test covering adapter binaries found only through additional GUI-missing PATH directories.

## Validation

- `xcodegen`

### DIFF
--- a/Alas/Sources/ACP/Adapter/ACPAdapterInstaller.swift
+++ b/Alas/Sources/ACP/Adapter/ACPAdapterInstaller.swift
@@ -15,6 +15,19 @@ enum ACPInstallError: LocalizedError {
     }
 }
 
+enum ACPProcessEnvironment {
+    static func augmented(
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #333, opened 2026-05-28
Persist worktree launch defaults per project

mrmans0n:
## Summary
- persist New Worktree open-after-create defaults per project
- restore project-specific launch surface when opening or switching projects in the dialog
- add coverage for config decoding, projects manager mutation, and launch default resolution

## Tests
- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test


mrmans0n:
@codex review

PR #338, opened 2026-05-28
Persist ACP composer drafts

mrmans0n:
## Summary
- Persist ACP composer drafts per session in the ACP SQLite store, including text and file mention chips.
- Restore drafts when ACP panes are recreated and after app restart.
- Defer clearing persisted drafts until async prompt send succeeds, preserving newer drafts typed while a prompt is in flight.

## Test Plan
- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/ACPComposerDraftTests -only-testing:AlasTests/ACPSessionStoreSchemaTests -only-testing:AlasTests/ACPSessionStoreCRUDTests -only-testing:AlasTests/ACPSessionManagerTests -only-testing:AlasTests/ACPSessionRunnerTests -only-testing:AlasTests/ACPComposerDraftBridgeTests`

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -71,6 +71,7 @@
 		162A6ABEAE92BB3AD1511567 /* AdvancedSettingsVisibilityTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1CD33324CA36118FCEA2BE2 /* AdvancedSettingsVisibilityTests.swift */; };
 		165300DDFFB445D8AAE7A5E5 /* MarkdownFileTypeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 19C327BB65463E878DDCDA5E /* MarkdownFileTypeTests.swift */; };
 		16F5212404948970EAFCFE62 /* MarkdownPreviewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0D4901C499EFE969EC36D116 /* MarkdownPreviewController.swift */; };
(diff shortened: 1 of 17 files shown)


mrmans0n:
Fixed the stale draft persistence path in 9d452703 by finalizing the submitted draft from the runner completion, guarded by the current in-memory draft so newer edits survive.

@codex review

mrmans0n:
Addressed the remounted-composer sync case in cfe79d6a: the mounted text view now syncs to persisted draft changes only when it still matches the last synced draft, so a successfully sent remounted prompt clears while local edits are preserved.

@codex review

mrmans0n:
Addressed the cancellation case in 143e320c: runner completions now treat a user-cancelled prompt error as an accepted submitted prompt for draft finalization, without setting `lastError`, and keep that cancellation state scoped by prompt id.

@codex review

mrmans0n:
Addressed the stale completion overlap in 1fb4d9d0: prompt completions only reset `streamingState` when they still own `activePromptID`, with coverage for cancel-then-immediate-resubmit.

@codex review

mrmans0n:
Addressed the stale cancelled completion finalization in b05a33bc: a cancelled prompt only reports accepted completion when no newer prompt is active, preventing an old identical submit from clearing the newer draft.

@codex review

mrmans0n:
Addressed the identical mid-turn draft case in 5ca078b0: submitted draft finalization now checks the captured draft revision as well as content, so a newer draft with the same text is not cleared by the earlier prompt completion.

@codex review

mrmans0n:
Addressed the stale cancelled prompt success path in aefb6878: both successful and failing completions suppress old prompt finalization when a newer prompt is active, with regression coverage for a cancelled prompt that later returns normally.

@codex review

mrmans0n:
Rebased on current `origin/main` and resolved the generated project conflicts in 976aa037. Local `xcodegen`, quiet build, and full test suite passed after the rebase.

@codex review

PR #339, opened 2026-05-28
fix(acp): render tool-call output in expanded chat cards

mrmans0n:
## Summary

- Tool cards in the ACP chat pane always showed "No output." when expanded, even for tools that clearly produced text. Root cause: we decoded the ACP `tool_call.content` field as `[ContentBlock]?` when the protocol actually defines it as `[ToolCallContent]?` — a tagged union wrapping a `ContentBlock`, a file `diff`, or a `terminal` reference. The wrapped `{"type": "content", ...}` shape every agent sends silently hit `ACPContentBlock`'s `default` branch and decoded to `.text("")`.
- Adds an `ACPToolCallContent` wrapper with the three real variants (plus an `.unknown` forward-compat bucket so a future spec variant doesn't get misdecoded as empty text again), switches `ACPToolCallPayload.content` and `ACPToolCallUpdate.content` to use it, and updates `ACPSession.flatten` to extract text per variant.
- Diff content renders as plain `--- path\n-old\n+new` lines (lossy but readable in the existing monospaced view); `terminal` is decoded but rendered as a placeholder — real terminal-output wiring is a separate follow-up.
- The pre-existing `session-update-tool-call.json` fixture was using a non-spec-compliant payload that only the buggy decoder accepted; fixed it to the real ACP wrapped shape and tightened the corresponding decode test so the regression can't slip back.

## Test plan

- [x] `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` — 2181/2181 pass
- [x] New decoder tests cover all four variants (`content`, `diff`, `terminal`, unknown)
- [x] New session-apply tests verify that wrapped text content reaches `tc.content` and that diff content renders to the exact expected string
- [x] Manual smoke: tool cards now show the tool's output when expanded

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		004A8BA5B9353F67AAA3AE83 /* NotificationService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2975A1465CA308E9766521F2 /* NotificationService.swift */; };
+		01568CB821D2E938C51BBCAB /* ACPHarnessBridgeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1A5A3E6B2F9AC2CBB1CDA78E /* ACPHarnessBridgeTests.swift */; };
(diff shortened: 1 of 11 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #345, opened 2026-05-28
feat(acp): O(1) streaming append via shared StreamingText buffer

mrmans0n:
ACPMessage.agent and .thought now carry a reference-typed StreamingText
buffer instead of an inline String. Appending a chunk mutates the
buffer in place (amortized O(1)) rather than reallocating the message
text and reassigning the @Published transcript array on every chunk
(which over the lifetime of a long agent reply was O(n²) in chars).

StreamingText is ObservableObject with @Published value. Row views
hold the buffer via @ObservedObject and re-render on append. Two
ACPMessage values that share the same buffer compare equal (identity
equality), so the transcript-array diff stays stable across streaming
growth — only the inner row re-renders.

Persistence is unchanged: codec encodes buf.value and decodes back
into a fresh buffer wrapper.


mrmans0n:

- 👉 #345
- #346

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #348, opened 2026-05-29
perf(acp): trim per-keystroke styling work in composer

mrmans0n:
ACPMarkdownLiveStyler.restyle ran on every keystroke and did two
things that compounded poorly with typing speed:

  1. NSFontManager.shared.font(...) was called twice per keystroke
     to compute the italic font. NSFontManager.shared takes a
     process-global lock; the italic font is constant. Base, bold,
     code, and italic fonts now live in `static let` properties.

  2. The whole text was re-styled on every keystroke — full
     enumerateAttribute scan + four regex.enumerateMatches passes,
     plus nested enumerateAttribute per match to skip chips. The
     regex patterns never cross a newline, so styling only needs to
     touch the lineRange of NSTextStorage.editedRange. Long drafts
     no longer pay O(n) per keystroke.

The styling output for any final text is unchanged.


mrmans0n:

- 👉 #348
- #349
- #350

mrmans0n:
@codex review

PR #349, opened 2026-05-29
perf(acp): fast-path composer draft extraction when no chips

mrmans0n:
Composer textDidChange called draft(from:) on every keystroke, which
walked the entire attributed string via enumerateAttribute to split
out chip mentions. The common case while typing is no chips at all,
so a single pre-scan checks for any .attachmentURI; when absent, the
draft is emitted as one .text segment without the per-range walk.


mrmans0n:

- 👉 #349
- #350

mrmans0n:
@codex review

PR #350, opened 2026-05-29
perf(acp): debounce SQLite composer draft writes

mrmans0n:
Every composer keystroke fired a synchronous SQLite UPSERT on the
main thread (JSONEncoder + db.exec). A typist at 8 chars/sec drove
8 writes per second per active composer.

In-memory state still updates instantly so cross-tab restore is
live. The SQLite write is now scheduled on a per-session Task that
sleeps 300ms then writes the LATEST in-memory draft. Further
keystrokes within the window cancel and re-schedule, so a typing
burst produces one write.

closeSession flushes the pending write before dropping the in-memory
session — otherwise a tab-switch-while-typing window can lose the
last ~300ms of input. deleteSession cancels pending writes since
the row is going away.

flushPendingDraftWrites() exposed for app-termination hooks and for
tests that need deterministic post-write reads.


mrmans0n:

- #349
- 👉 #350

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #351, opened 2026-05-29
fix(terminal): strip ZMX_SESSION from inherited env to stop cross-pane re-attach

mrmans0n:
## Summary
- When Alas itself is launched from inside a zmx-attached terminal, the parent env carries `ZMX_SESSION` (and `ZMX_SESSION_PREFIX`) through every pane we spawn. Any subsequent `zmx attach <other>` from that shell — directly, via an agent CLI, a completion hook, anything — falls into zmx's `switchSesh` path (it reads `ZMX_SESSION` before the CLI argument and forwards a `Switch` IPC), silently re-binding the active pane to a different daemon's session. End state: one tab is piped into another tab's shell, and the original daemon shuts down once its last client leaves (its on-disk socket file disappears).
- Add `ZMX_SESSION` and `ZMX_SESSION_PREFIX` to `EnvBuilder.strippedKeys` so every pane gets only its own scoped name.
- See [zmx#151](https://github.com/neurosnap/zmx/issues/151) for the upstream behaviour; this PR closes the leak from the Alas side.

## Test plan
- [x] New `EnvBuilderTests.stripsZmxSessionVarsFromInheritedParent` mirrors the existing TERM-stripping coverage.
- [x] `xcodebuild ... -only-testing AlasTests/EnvBuilderTests test` — all 9 tests pass locally.
- [ ] CI green.
- [ ] Smoke: open Alas from a zmx-attached terminal, spawn a tab, run `zmx attach <other>` — original pane stays bound to its own session.

### DIFF
--- a/Alas/Sources/Terminal/EnvBuilder.swift
+++ b/Alas/Sources/Terminal/EnvBuilder.swift
@@ -9,8 +9,12 @@ enum EnvBuilder {
         // terminal, or because we're building env for a sibling pane), passing
         // it through means a subsequent `zmx attach <other>` from anywhere in
         // the new shell falls into zmx's switchSesh path and re-attaches the
-        // pane to a different session. See zmx#151.
-        "ZMX_SESSION", "ZMX_SESSION_PREFIX",
+        // pane to a different session. See zmx#151. Note: ZMX_SESSION_PREFIX
+        // is intentionally NOT stripped — it's a user-facing zmx config knob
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #354, opened 2026-05-29
feat(acp): off-main session hydration + windowed transcript render

mrmans0n:
## Summary

Opening a worktree whose active tab is an ACP session — or switching to an ACP tab for the first time in this app launch — used to freeze the UI while `ACPSessionManager` loaded the row, read every persisted message from SQLite, JSON-decoded each one with a fresh decoder, and rendered the entire historical transcript on the first paint. The freeze grew linearly with transcript length.

This PR moves the heavy work off the main thread and bounds first-paint cost on long sessions.

**Hydration off main:**
- New `ACPSessionHydrator` actor owns its own `SQLiteDatabase` handle and runs the whole hydration pass (`loadMessages`, `loadQueue`, `loadComposerDraft`, `upsertSession` touch, `recentSessions` refresh) off main.
- New `ACPMessageWire` is a `Sendable` mirror of `ACPMessage` that crosses the actor boundary; the main hop converts wire variants into full `ACPMessage`s and is the only place `StreamingText` gets allocated.
- `ACPSession` gains a `HydrationState` flag; `ACPTabView` reuses the existing `ACPConnectingPlaceholder` while loading, runs hydrate before attach, and surfaces a Retry banner on failure.
- `ACPSessionManager.openSession` splits into a sync `placeholderSession` (cache hit O(1); cache miss is a single indexed `loadSession`) plus an async `hydrateIfNeeded` that coalesces concurrent callers via an in-flight map.

**Windowed transcript render:**
- `ACPTranscript` gains `visibleHead` + `tailWindow = 30`. After hydration applies, the head is reset to `max(0, count - 30)` so long sessions paint immediately.
- `ACPMessageList` slices to `messages[visibleHead...]` and shows an "Earlier messages…" capsule above the rendered range. Scrolling the marker within ~200pt of the top steps back another 30 rows (debounced 300ms).

**Tests:** 30 new tests across 5 new suites cover the wire round-trip, hydration state, transcript window math, hydrator I/O (happy path / missing session / malformed payload skip / `lastOpenedAt` touch), and manager hydration paths (placeholder semantics, `.ready`/`.failed` transitions, window reset, concurrent call coalescing).

## Test plan

- [x] Full ACP suite: 71/71 green across the 9 affected suites.
- [x] Manual: cold open of a worktree with an active ACP tab pointing at a long-history session shows the placeholder immediately, then the windowed transcript, with no perceptible main-thread stall.
- [x] Manual: switching between ACP tabs is instant on subsequent activations (hydration cache hit).
- [x] Manual: new session creation unchanged — placeholder during `attach`, composer accepts input as soon as setup succeeds.
- [ ] CI green.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -205,6 +205,7 @@
 		3FCC4EDA31A96BE121F45D44 /* ACPComposerPlaceholderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85E54E9F53ECFE9C205788F1 /* ACPComposerPlaceholderTests.swift */; };
 		3FE3D85D5E833DE3A91FD444 /* CopilotInstaller.swift in Sources */ = {isa = PBXBuildFile; fileRef = D64C0B7363363C77718B446C /* CopilotInstaller.swift */; };
 		40450134EBF5CCD5CD667327 /* ChangeSummaryVisibility.swift in Sources */ = {isa = PBXBuildFile; fileRef = 524926F2D16B1AD83FC234AB /* ChangeSummaryVisibility.swift */; };
+		4058A537B8CAE76FE7EFE184 /* ACPSessionHydrationStateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1DC1B975662D135B148443BC /* ACPSessionHydrationStateTests.swift */; };
 		40D9FD39D36ADE353B9971E8 /* MergeRegionVisualLayoutTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 256257983A4B1E1E674DF5B8 /* MergeRegionVisualLayoutTests.swift */; };
 		412F8D73E3E5CD7F57CD799A /* CompletionWindowController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31F3A80227DD4A52C4752722 /* CompletionWindowController.swift */; };
 		4153DAD75F27395D2D385BED /* LanguageServerAvailabilityTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A8CC42BD54479E057902D392 /* LanguageServerAvailabilityTests.swift */; };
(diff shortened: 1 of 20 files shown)


mrmans0n:
@codex review — addressed the draft-overwrite race by guarding the restore on `session.composerDraft.isEmpty` and added a regression test (`hydratePreservesInProgressDraft`).

mrmans0n:
@codex review — addressed the two P2s:
- `hydrateIfNeeded` now recurses after awaiting an in-flight task, and `runHydration` re-checks `sessions[id] === session` before applying. A close+reopen during in-flight hydration now drives the reopened placeholder to `.ready` (test: `hydrateAfterCloseReopen`).
- Draft-restore guard switched from `composerDraft.isEmpty` to `composerDraftRevision == 0` so a deliberate clear is preserved (test: `hydrateRespectsDeliberateClear`).

mrmans0n:
@codex review — addressed both new P2s:
- `hydrateIfNeeded`: the spawned task now clears its own `inFlightHydrations` entry before returning, so a waiter that resumes from `await existing.value` always sees a clean slate before recursing. No more spinning on the completed task.
- `ACPMessageList`: head-step guard now requires `frame.maxY > 0` in addition to `minY < threshold`. When the marker is parked far above the viewport (e.g. initial restore-to-tail on a long transcript), the preference fire is ignored.

mrmans0n:
@codex review — addressed the three new P2s:
- **Hydrator touch resurrecting rows**: replaced `upsertSession(touched)` with a narrow `touchLastOpenedAt(id:, at:)` UPDATE that no-ops when the row is gone and never rewrites `archived`. Same change applied to the sync test fallback. New tests: `deletedSessionStaysDeleted`, `archivedSessionStaysArchived`.
- **Title overwrite during hydration**: `applyHydration` no longer assigns `session.title` — `placeholderSession` already seeded it from the row, so any rename made during the loading window survives. New test: `hydratePreservesInProgressTitleEdit`.
- **Retry never attaches**: extracted `hydrateAndAttach()` from `.task(id:)` and routed the failure-banner Retry through it, so a successful retry continues into `manager.attach` exactly like the initial mount.

mrmans0n:
@codex review — tightened the stale-task guards in `hydrateAndAttach`. Before attaching we now require: (a) the task isn't cancelled, (b) `manager.sessions[sessionId] === session` (no close+reopen swap), and (c) `hydrationState == .ready` (catches the case where our result was intentionally dropped because the visible placeholder changed under us).

PR #359, opened 2026-05-29
feat(acp): nudge users when an installed adapter has updates

mrmans0n:
## Summary

- When an installed npm-managed ACP adapter (`claude`, `codex`, `pi`) has a newer version on the npm registry, surface a one-click **Update** banner in the session view — reusing the same UX pattern as the existing first-install nudge.
- Lazy on session attach: after `.ready`, an actor-backed store consults `npm outdated -g <pkg> --json` (5s timeout) and caches per-agent (24h success, 30m failure). A pure decider routes install/update/none, install always wins.
- X-dismissing persists the dismissed `latest` so the banner reappears only when npm publishes something newer. Failures (offline, missing npm, timeout) are silent.

## Architecture

- New: `AdapterUpdateState`, `ACPAdapterVersionChecker` (runner-injected), `ACPAdapterUpdateStore` (actor, TTL + dismissal + in-flight coalescing), `ACPAdapterUpdateBannerDecider` (pure).
- Extended: `ACPSetupNudgeBanner` gains a `Mode` (`.install` / `.update(current, latest)`) with mode-aware copy. Both modes share the same `npm install -g <pkg>` runner — no separate update command.
- Wiring lives in `ACPTabView.hydrateAndAttach` (lazy check after attach) and `reattachAfterUpdate` (clears cache + re-checks after a successful update).

## Test plan

- [ ] `xcodebuild test` — ~50 new tests across 5 files (version-check parsing + timeout, codable round-trip, TTL/dismissal/coalescing/persistence, banner copy by mode, decider precedence, launch-spec helper).
- [ ] Manual: downgrade `@agentclientprotocol/claude-agent-acp` via `npm install -g <pkg>@<older>`, reopen a Claude ACP session — Update banner should appear with both versions. Click Update → reconnects. Dismiss → stays gone; re-bump to a still-newer version → banner reappears.
- [ ] Manual: with no adapter installed, install nudge still works unchanged.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -29,6 +29,7 @@
 		065F01BB553C415040414D37 /* ImageDiffSwipeView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 802EF6C18E2E27F68F6B770E /* ImageDiffSwipeView.swift */; };
 		065F1A7FCF81FF29BAEF5A86 /* ScopeRow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4D131C3F1E4BE68A850097A4 /* ScopeRow.swift */; };
 		06AD46194E45984A1297D928 /* CommitHeaderViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E43F9F04AF6A6FF0A4D328F4 /* CommitHeaderViewTests.swift */; };
+		07BB2B329992BA535053BB80 /* ACPSetupNudgeBannerModeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0AB6FA9F258818E8B464CB0A /* ACPSetupNudgeBannerModeTests.swift */; };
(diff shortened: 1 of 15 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
Codex P2 + P3 addressed in bed493f6:

**P2** — `defaultRunner` now wires `withTaskCancellationHandler` to `Process.terminationHandler` via a continuation; timeout-driven cancellation calls `proc.terminate()`, which lets the per-agent in-flight slot in the store actually clear at the bound.

**P3** — `parse` returns `.upToDate` when JSON parses but the package key is absent (e.g. `{}`), so up-to-date adapters get the 24h success TTL instead of the 30min failure TTL. New test `emptyJsonObjectIsUpToDate` plus flipped `missingPackageEntry` expectation.

@codex review

mrmans0n:
Codex P2 follow-up addressed in 812da968: parser now only treats an empty `{}` root as up-to-date; any other JSON shape without our package key (e.g. `{"error":...}` on registry/auth failures) stays `.unknown` so the 30m failure TTL applies. New `errorPayloadIsUnknown` test pins the behavior; flipped `missingPackageEntry` back to expect `.unknown` for the unrealistic other-package case.

@codex review

PR #362, opened 2026-05-29
perf(acp): in-app memory accounting + capped retain surfaces

mrmans0n:
## Summary
- **Motivation**: under a 6-worktree / 2–3 ACP-pane daily workload, Alas grew unbounded to ~8 GB and was OOM-killed. Smoking gun: `ACPTranscript.markdownCaches` retained every streaming reply's full text + parsed AST forever (`dropMarkdownCache` / `resetMarkdownCaches` were defined but never called). Tool-call `content` (whole file reads, command output) also lived in memory indefinitely.
- **What this lands**: in-app per-subsystem memory accounting (debug-only) + four targeted caps on the retain surfaces flagged by the spec, with the test infra to verify them.

### Diagnostics (debug-only)
- `MemoryDiagnostics`: pull-model rollup of per-session transcript bytes, markdown-cache bytes, runner counts, plus kernel `phys_footprint` via `task_info(TASK_VM_INFO)`.
- 30s tick emits a one-line `mem: phys=… acp.tx=… acp.md=… term=… sessions=… runners=… unattributed=…` via `os.Logger`.
- ⌘⇧⌃M opens a "Memory report…" window showing per-session breakdown sorted by transcript bytes desc; "Refresh now" pulls a fresh sample.
- Entire layer compiled out of release builds (no `task_info` syscall, no metering types).

### Retain surface caps
1. **Markdown cache eviction.** `ACPTranscript.setVisibleHead` is the single funnel that advances the render window head AND drops markdown caches for messages dropped below it. Retreating writes leave caches intact. `closeSession` / `deleteSession` also reset all caches.
2. **Tool-call content truncation.** When the window advances, completed (non-`in_progress` / non-`pending`) tool calls whose index drops below the head are truncated in-memory to 4 KB; `isContentTruncated` is excluded from `Equatable` / `Hashable` / `Codable` (in-memory only). Full content stays in SQLite and is refetched via `ACPSessionManager.reloadFullToolCallContent` when the user expands an old card. `ACPToolCallCard` resets its `@State` cache on `toolCallId` change so view recycling can't show stale content.
3. **Idle session eviction.** `ACPSessionManager` tracks a per-session UI refcount via `retainSession` / `releaseSession`. When refcount hits zero AND the session's `agentState` is not `.spawning` / `.ready`, the cached `ACPSession` is dropped from the in-memory map. Re-opening re-hydrates through the existing placeholder + hydrate path. Tabs and the history sidebar retain on materialization, release on dismiss.
4. **Runner audit.** `ACPSessionRunner` inspected end-to-end — all unstructured `Task` blocks use `[weak self]` and are cancelled in `stop()`; the three stored `for await` loops exit on cancel; no Combine sinks; no closures stored on `ACPConnection` that point back at the runner. Clean.

### Out of scope (deferred per the spec)
- Soft cap on individual `StreamingText` buffers (only matters if a single agent reply exceeds a few MB; instrumentation will surface this).
- Lower terminal scrollback default (already bounded at 10K lines/pane; only adjust if metering shows it's a meaningful contributor under the workload).

## Test plan
- [x] `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` — full suite passes (2503 tests, 0 failures)
- [x] 17 new Swift Testing cases covering: snapshot formatting & unattributed clamp, `phys_footprint` probe non-zero + grows-after-allocation, per-session byte sums per message kind, markdown cache eviction including idempotence + retreat-keeps + reset-clears, tool-call truncation including live/pending skip + equality round-trip + flag exclusion, refcount eviction including attached-stickiness + unknown-id no-op + closeSession override
- [x] Release build (`-configuration Release`) compiles cleanly — confirms the DEBUG gate is correct
- [ ] Run the user's real-workload reproduction (6 worktrees, 2–3 ACP panes, long sessions over several hours): `phys_footprint` should stabilise under ~2 GB; `acp.tx` and `acp.md` should not trend upward across hours; `unattributed` bucket should not grow

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -37,6 +37,7 @@
 		08DBBE562CC62A888617AC09 /* SidebarHeaderViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 25DBE942154BD154F2D10B9D /* SidebarHeaderViewTests.swift */; };
 		08DF3AC4BA50775940371445 /* ACPSessionStoreTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0A443685B2A3322208D51AE /* ACPSessionStoreTests.swift */; };
 		095A19B85DFA5EBA5A24786D /* AdvancedPane.swift in Sources */ = {isa = PBXBuildFile; fileRef = E16C4F54FB3E390F80208DA1 /* AdvancedPane.swift */; };
+		095E7D98616FB77A9F8BDFD9 /* ACPSessionManagerRetentionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6D1AAAE396590865426C592A /* ACPSessionManagerRetentionTests.swift */; };
 		0A480D9D7CE5BB837D7BCA13 /* GitService+Staging.swift in Sources */ = {isa = PBXBuildFile; fileRef = D2F90D03073708AA01531533 /* GitService+Staging.swift */; };
 		0B11B8038CBA3A6F05E34895 /* GatekeeperAssessorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 137E9C84D7F32E747F0AF4C0 /* GatekeeperAssessorTests.swift */; };
 		0BA997A903C1DC2A41D64E22 /* AgentRunner.swift in Sources */ = {isa = PBXBuildFile; fileRef = A919B8770029DBE7FCFF0EE2 /* AgentRunner.swift */; };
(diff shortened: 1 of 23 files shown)


mrmans0n:
@codex review

Addressed in 94efd3fa: `detach` now calls `evictIfIdle(id:)` after the runner shutdown completes, so a tab closed while attached evicts as soon as the async detach flips `agentState` out of `.ready`. Added a regression test (`detachEvictsAfterAsyncRelease`) that simulates the SwiftUI `.onDisappear` race: refcount drops to zero while still `.ready` (eviction skipped), then `detach` runs and the session is dropped.

mrmans0n:
@codex review

Addressed in d3585003: `evictIfIdle` now calls `flushPendingDraftWrite(for: id)` before dropping the session, so the debounced 300ms composer-draft timer can't lose its tail when the session is evicted while a write is still pending. Added a regression test (`evictionFlushesDraft`) that persists a draft via the debounced path, evicts, and asserts the draft is in the store.

PR #370, opened 2026-05-29
fix(git): force LC_ALL=C so worktree-remove works in non-English locales

mrmans0n:
## Summary
- `WorktreeService` parses git stderr (English-only substrings) to detect submodule / dirty / missing-LFS errors and trigger the auto-force-remove path. Under a non-English shell locale git emits translated messages, none of the matchers fire, and the user sees the raw error instead of the intended recovery — reported as `fatal: árboles de trabajo conteniendo submódulos no pueden ser movidos o eliminados` when trying to delete a worktree containing submodules.
- Fix: pin `LC_ALL=C` in `Process.gitEnv()` so every git invocation emits parseable English regardless of the user's environment. POSIX guarantees `LC_ALL` overrides `LANG` and `LC_MESSAGES`.

## Test plan
- [x] Shell-level repro: `LANG=es_ES.UTF-8 git worktree remove …` → Spanish; same with `LC_ALL=C` → English (exact string the matcher checks)
- [x] `ProcessGitTests.gitEnvForcesCLocale` — asserts `gitEnv()["LC_ALL"] == "C"`
- [x] `ProcessGitTests.gitEmitsEnglishMessagesUnderForeignLocale` — runs `git help -a` with an injected `LANG=es_ES.UTF-8` and verifies English output
- [x] `ProcessGitTests.gitEnvPreservesAllParentVariables` updated to include `LC_ALL` in the override exception list
- [x] All 44 affected tests pass (`ProcessGitTests` + `WorktreeServiceTests`, incl. existing submodule scenarios)

### DIFF
--- a/Alas/Sources/Git/Process+Git.swift
+++ b/Alas/Sources/Git/Process+Git.swift
@@ -168,6 +168,12 @@ extension Process {
     static func gitEnv() -> [String: String] {
         var env = ProcessInfo.processInfo.environment
         env["GIT_OPTIONAL_LOCKS"] = "0"
+        // Force git to emit messages in English. We parse stderr in several
+        // places (e.g. WorktreeService submodule/dirty/LFS matchers) and those
+        // matchers are English-only; a user with `LANG=es_ES.UTF-8` would
+        // otherwise hit translated errors that no matcher recognizes. POSIX:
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #375, opened 2026-05-30
fix(ACP): tail scroll pause threshold

mrmans0n:
## Summary

Fixes ACP transcript tail-follow behavior so auto-scroll is not disabled by tiny upward hops near the bottom of the pane.

## Root Cause

The scroll classifier paused tail-following on any upward offset change larger than sub-pixel jitter once the viewport was outside the 36pt bottom resume tolerance. That made small trackpad/layout hops near the tail latch turn auto-scroll off while agents continued streaming.

## Changes

- Add a separate 160pt pause tolerance before upward movement can disable tail-following.
- Keep the existing 36pt bottom tolerance for resuming tail-following.
- Preserve restore-time protection so programmatic restore movement is ignored, while meaningful user upward movement can still pause.
- Add classifier tests for near-tail hops, meaningful-away upward movement, restore-time near-tail input, and pause boundary behavior.

## Validation

- `xcodegen` passed.
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build` passed.
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/ACPScrollDirectionClassifierTests` passed: 23 tests.
- GitHub Actions `Build / build-test` passed on commit `15125ff`.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPScrollDirectionClassifier.swift
+++ b/Alas/Sources/ACP/UI/ACPScrollDirectionClassifier.swift
@@ -11,12 +11,17 @@ enum ACPScrollDecision: Equatable {
 /// AppKit/SwiftUI so it can be exercised in unit tests.
 enum ACPScrollDirectionClassifier {
     /// Sub-pixel jitter is ignored. 0.5pt comfortably covers Retina rounding
-    /// noise; anything larger is treated as user intent.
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #380, opened 2026-05-30
fix(acp): lossless queued-prompt edit-restore + atomic edit-take

mrmans0n:
## What & why

Follow-up to #378 ("A"), addressing the two threads of Codex feedback that were deferred there.

### 1. Lossless queued-prompt round-trip
The composer holds structured state (`ACPComposerDraft.segments`: `.text`/`.mention`). On submit it's flattened to wire `blocks` (`[.text(wholeString), .resourceLink…]`), which loses the chip-vs-literal distinction — so editing a queued prompt had to invert it with a heuristic forward-scan (`ACPComposerDraft(blocks:)`) that could misclaim a literal `@name` the user typed. In #378 I accepted that as a limitation, noting a lossless fix needs to stop relying on inverting the lossy serialization.

**Fix:** capture the structured draft on the queued item — `QueuedPrompt.draft` (optional, persisted, tolerant decode) — and restore from it exactly via `restorableDraft`. The wire `blocks` are **unchanged**; `draft` is additive and read only on edit-restore. The heuristic survives solely as a fallback for draft-less items (pre-upgrade persisted, recovery-path, rolled-back sends).

### 2. Queue-edit race (the unaddressed thread on #378)
`removeFromQueue(id:)` no-ops once an item is `.sending`, but `onQueueEdit` restored into the composer **then** removed — so a flusher promoting the item between render and click could duplicate an in-flight prompt.

**Fix:** `ACPSession.takeForEditing(id:)` atomically removes-and-returns the restorable draft **only if still `.pending`** (nil otherwise). `@MainActor`-confined, so the check/remove can't interleave. The UI restores only on non-nil.

## How
- `QueuedPrompt.draft: ACPComposerDraft?` + `restorableDraft` (prefers draft, else heuristic).
- `ACPSession`: `enqueue(blocks:draft:)`, atomic `takeForEditing(id:)`, `editQueueItem` clears the stale draft when blocks change.
- Thread optional `draft` through `ACPSessionManager.submit` / `enqueueWhileRecovering` / `ACPSessionRunner.send` to the `.enqueue` route (ignored on `sendNow`/`steer`).
- `ACPTabView`: `onQueueEdit` → `takeForEditing`; submit forwards `draft`.

## Tests
- `QueuedPrompt`: draft codable round-trip, legacy JSON (no draft key) → nil, `restorableDraft` fallback.
- `ACPSession` queue API: `enqueue(draft:)`, `takeForEditing` for pending/sending/unknown/no-draft, the **lossless `@name`-collision** proof (heuristic misclaims vs stored-draft exact), `editQueueItem` clears draft.
- 44 tests across the ACP session/composer/manager/store suites pass; app target builds.

## Test plan (manual UI)
- [ ] Mid-text mention round-trips on edit (chip stays a chip, no duplicate)
- [ ] Literal `@name` + a chip for a same-named file both restore correctly
- [ ] Clicking edit on a `.sending` item does nothing (no duplication)
- [ ] Non-empty composer: queued item appends after typed text without clobbering
- [ ] Mention restores correctly after app relaunch (persistence path)

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSession.swift
+++ b/Alas/Sources/ACP/Session/ACPSession.swift
@@ -231,8 +231,8 @@ final class ACPSession: ObservableObject, Identifiable {
     /// Append a new pending item to the tail of the queue. Used by the
     /// runner when the user submits while the agent is busy (or while
     /// the queue is already non-empty — see ACPSubmitRoute).
-    func enqueue(blocks: [ACPContentBlock]) {
-        queue.append(QueuedPrompt(blocks: blocks))
+    func enqueue(blocks: [ACPContentBlock], draft: ACPComposerDraft? = nil) {
+        queue.append(QueuedPrompt(blocks: blocks, draft: draft))
     }
 
     /// Remove a specific item by id. The drag-handle X on the bubble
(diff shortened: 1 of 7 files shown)


mrmans0n:
@codex review

PR #383, opened 2026-05-30
fix(acp): move focus outline to composer

mrmans0n:
## Summary
- suppress the ACP pane-level focus effect while keeping keyboard focusability
- use the green add token for the focused ACP composer outline
- leave existing composer button focus styling unchanged

## Validation
- xcodegen
- git diff --check

Full xcodebuild test was attempted during implementation but could not complete because the system volume ran out of space, then Xcode test orchestration stalled after moving DerivedData/TMPDIR to the external volume.


mrmans0n:
@codex review

PR #405, opened 2026-05-31
fix(ACP): composer draft observation jank

mrmans0n:
## Summary

- Move rapidly-changing ACP composer draft state into a narrow `ACPComposerState` observable.
- Keep existing `ACPSession.composerDraft` and revision accessors for persistence and submit flows while avoiding broad session invalidation on every keystroke.
- Wire the composer input field to observe composer state directly and add regression coverage for the observation boundary.

## Root cause

`ACPSession.composerDraft` was `@Published`, so every composer keystroke published the whole session. With a long ACP transcript open, that could invalidate the transcript pane and scroll/markdown bookkeeping while typing, causing app-wide jank and composer lag.

## Validation

- `rtk xcodegen`

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSession.swift
+++ b/Alas/Sources/ACP/Session/ACPSession.swift
@@ -1,6 +1,21 @@
 import Foundation
 import Combine
 
+/// Narrow observable holder for the rapidly-mutating composer draft.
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

PR #406, opened 2026-05-31
fix(ACP): suppress ACP load replay after hydration

mrmans0n:
## Summary

- Suppress buffered `session/load` replay updates when reopening an already-hydrated ACP session.
- Keep Alas's persisted SQLite transcript authoritative so user bubbles and image attachments restore once.
- Add regression coverage for replaying load, fresh sessions, and non-replaying agents with queued prompts.

## Root Cause

Reopening an ACP session hydrated the transcript from Alas's SQLite store, then consumed the agent's full replay from `session/load` as normal live updates. Those replay updates were appended after the hydrated transcript, duplicating the conversation.

## Validation

- `rtk xcodegen`

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSessionManager.swift
+++ b/Alas/Sources/ACP/Session/ACPSessionManager.swift
@@ -583,8 +583,30 @@ extension ACPSessionManager {
                 stderrBuffer.append(data)
             }
         }
+        var startedRunner: ACPSessionRunner?
         do {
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #411, opened 2026-05-31
fix(acp): dedupe agent discovery path entries

mrmans0n:
## Summary

- continue the ACP provider auto-discovery plan in the current Swift detector
- dedupe absolute PATH directories before probing agent binaries
- add regression coverage that preserves first-seen PATH order

## Validation

- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/AgentDetectorTests test`\n- `rtk xcodegen`\n- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build`\n- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test`

### DIFF
--- a/Alas/Sources/Agents/AgentDetector.swift
+++ b/Alas/Sources/Agents/AgentDetector.swift
@@ -37,16 +37,17 @@ enum AgentDetector {
     }
 
     static func executableSearchDirectories(path: String) -> [String] {
-        path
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #412, opened 2026-05-31
Fix merge conflict gutter alignment

mrmans0n:
## Summary
- Split merge conflict RESULT ranges into LOCAL and REMOTE segments.
- Connect each action gutter to the matching RESULT segment instead of the full conflict block.
- Add regression coverage for normal and BASE-visible conflict layouts.

## Validation
- `xcodegen`


mrmans0n:
@codex review

PR #415, opened 2026-05-31
fix(acp): reduce long session scaling work

mrmans0n:
## Summary

This reduces ACP long-session scaling work in the hot update path.

- Avoids reloading and materializing every persisted message just to decide whether an ACP update should append or update a row.
- Tracks persisted message count in the runner and adds a narrow SQLite count helper.
- Adds a tool-call id to message-index map so tool-call updates do not scan the whole transcript every time.
- Advances the live render window while the transcript is following the tail, so long sessions do not keep every historical row rendered until a reopen.
- Adds regression coverage for cheap message counting and live tail-window advancement.

## Validation

- `rtk xcodegen`
- Focused ACP tests for session, store, runner, hydration, attach/restore, and tool-call truncation

Full `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` currently fails in unrelated `BaseBranchSelectorTests.branchListHeightMatchesVisibleRows` exact `CGFloat` equality assertions; the ACP-focused suites pass.

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSession.swift
+++ b/Alas/Sources/ACP/Session/ACPSession.swift
@@ -93,6 +93,7 @@ final class ACPSession: ObservableObject, Identifiable {
     var remoteSessionId: String?
     @Published var queue: [QueuedPrompt] = []
     @Published var steerUndo: SteerUndoState?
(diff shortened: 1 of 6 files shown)


mrmans0n:
@codex review

PR #420, opened 2026-06-01
fix(acp): tail-first transcript hydration

mrmans0n:
## Summary

Long ACP sessions froze the UI on every tab activation because `applyHydration` ran an O(N) main-actor pass to wrap every persisted message in `StreamingText` (a `@MainActor` class) before the first paint could land — even though SwiftUI only drew the tail window once the data finally arrived.

Now hydration is split in two:

1. **Tail-first paint.** The last `tailWindow` messages (30) are converted on the synchronous path and applied with `visibleHead = 0`, so the tab paints immediately and the user sees the conversation they expect.
2. **Background backfill.** Older wires are materialised on a follow-up `Task` that yields between 100-message batches and prepends the result in one transcript mutation. `visibleHead` is shifted by the prepended count so the visible tail stays anchored — no layout jump.

The new `awaitBackfill(id:)` API lets tests observe the fully-materialised transcript; production UI doesn't need it. Pending backfills are cancelled in `closeSession`, `deleteSession`, and `evictIfIdle` so a re-hydration (or eviction) can't race with a stale prepend.

## Test plan

- [x] `ACPSessionManagerHydrationTests` — 16/16, including new `hydrateAppliesTailFirst` (asserts tail-only after hydrate; full transcript after `awaitBackfill`) and `hydrateShortTranscriptHasNoBackfill` (one-pass path for short sessions).
- [x] Existing `hydrateResetsWindow` updated to await backfill before asserting the final `visibleHead == count - tailWindow`.
- [x] Full ACP test surface: 114/114 across `ACPSessionTests`, `ACPSessionManagerAttachRestoreTests`, `ACPSessionManagerTests`, `ACPSessionManagerRetentionTests`, `ACPSessionRecoveryTests`, `ACPSessionRecoveryIntegrationTests`, `ACPSessionHydratorTests`, `ACPTranscriptWindowTests`, `ACPTranscriptMarkdownCacheEvictionTests`, `ACPSessionRunnerTests`.
- [ ] Manual: open a long ACP session, switch tabs, confirm no perceptible freeze.

## Follow-ups (not in this PR)

- Persist parsed-markdown snapshot so first paint doesn't re-parse markdown for the tail-window agent messages. This is the next planned change on top of this branch.

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSession.swift
+++ b/Alas/Sources/ACP/Session/ACPSession.swift
@@ -295,6 +295,26 @@ final class ACPSession: ObservableObject, Identifiable {
         rebuildToolCallIndices()
     }
 
+    /// Insert `older` at the head of the transcript and shift `visibleHead`
+    /// forward by `older.count` so the visible tail window stays anchored to
+    /// the same messages. Used by the post-hydration backfill that prepends
+    /// pre-tail messages after the UI has already painted the tail.
+    func prependTranscriptMessages(_ older: [ACPMessage]) {
+        guard !older.isEmpty else { return }
+        transcript.messages.insert(contentsOf: older, at: 0)
+        // Rebuild tool-call indices because every prior entry's index just
+        // shifted by `older.count`. Cheaper than offsetting in place: the
+        // cache is dictionary-typed, so a full rebuild is O(N) and avoids
+        // any chance of drift if a streaming update lands mid-shift.
+        rebuildToolCallIndices()
+        // Keep the visible window anchored to the tail the user is already
+        // looking at. Bypass `setVisibleHead` — it strips markdown caches
+        // for messages below the new head, which is exactly what we want
+        // to *avoid* here (older messages weren't visible before the shift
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

mrmans0n:
Thanks @codex — pushed 79144fd that gates runner attach on `awaitBackfill` (and same for the markdown-export path), with a regression test that asserts the setup evaluator observes the fully-materialised transcript. @codex review

mrmans0n:
Pushed 2fd8ab6 addressing the P2: backfill cleanup now compares against a captured task handle, so a cancelled-then-superseded task can't wipe the newer task's slot in `inFlightBackfills`. @codex review

mrmans0n:
Pushed 97f15ba addressing the recovery-warning P2: `applyHydration` now computes `canSendTranscript` against the full wire list via a `wireMessagesHaveConversation` helper, so a long session whose tail is non-conversation messages still surfaces the warning correctly. Regression test asserts the buried-user-prompt case. @codex review

PR #421, opened 2026-06-01
Preserve user env for ACP auth terminals

mrmans0n:
## Summary
- force ACP auth terminal launches to inherit the parent environment even when the normal terminal setting disables inheritance
- keep the auth-only startup script and agent-marker sanitization behavior intact
- add regression coverage for auth launch inheritance

## Verification
- `rtk swiftformat Alas/Sources/App/AppState.swift AlasTests/AgentTerminalLaunchTests.swift --lint --reporter github-actions-log`
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/AgentTerminalLaunchTests -quiet test`

### DIFF
--- a/Alas/Sources/App/AppState.swift
+++ b/Alas/Sources/App/AppState.swift
@@ -821,6 +821,7 @@ final class AppState {
             for: worktree,
             startupScriptSuffix: startupScriptSuffix,
             includeUserStartupScript: false,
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #426, opened 2026-06-01
feat(changes): toggle to track upstream branch in Commits section

mrmans0n:
## Summary

- Adds a Settings toggle at `Changes › Commits section › Track upstream branch` (default **off**). When off, the Commits section in the Changes tab always compares HEAD against `worktrees.baseBranch` (preferring local mainline, falling back to `origin/<base>` — matching the existing `GitService.commitsAhead` cascade semantics for `ignoreUpstream`). When on, the legacy upstream-aware cascade is used: once you push, the section only lists unpushed commits.
- The "behind upstream" chip is unchanged — it uses an independent probe (`resolveUpstreamRef`) and remains useful regardless of this toggle.
- Same toggle wires into the AI commit-message generator's "nearby commits" context fetch in `CommitEditorTabView`, so both surfaces stay consistent.
- Existing user configs decode the missing key as `false` and pick up the new boring default on first launch.

## Behavior matrix

| Branch state | Toggle ON (legacy) | Toggle OFF (new default) |
|---|---|---|
| Unpushed, ahead of `main` | lists commits vs `origin/main` | lists commits vs `main`/`origin/main` |
| Pushed, even with upstream | **empty list** | lists commits vs `main`/`origin/main` |
| Pushed, behind upstream | lists upstream-only commits as "yours" | lists commits vs base |
| On `main` itself | empty | empty |

## Known follow-up (not blocking)

The design spec specified "prefer `origin/<base>`, fall back to local" for the toggle-off cascade, but `GitService.commitsAhead(ignoreUpstream: true)` actually prefers local when both exist (locked in by the existing `explicitSimpleBasePrefersLocalBranchOverOrigin` test). In practice this is fine — local `main` usually matches `origin/main` closely and the "N behind" chip already surfaces divergence — but worth a follow-up if we want strict origin-first semantics for the toggle-off case without changing user-override behavior.

## Test plan

- [x] Unit tests for config decode default + round-trip (`AppConfigChangesTests`).
- [x] Unit test for `RightPaneStore` refresh-on-toggle path (`storeRefreshesWhenTrackUpstreamForCommitsToggles`).
- [x] Existing upstream-cascade tests opted into `trackUpstreamForCommits = true` to remain valid.
- [ ] Manual: pushed feature branch with toggle off → Commits section lists everything since base.
- [ ] Manual: same branch, toggle on → Commits section empties (or shows only unpushed commits).
- [ ] Manual: flip toggle while a worktree is open → list updates without restarting the worktree.
- [ ] Manual: "N behind <ref>" chips behave the same regardless of toggle.

### DIFF
--- a/Alas/Sources/App/AppState.swift
+++ b/Alas/Sources/App/AppState.swift
@@ -1945,7 +1945,11 @@ final class AppState {
         guard let worktree = worktree(withId: worktreeId) else { return }
         config.rightPaneVisible = true
         _ = saveConfig()
-        let rps = rightPaneStore.state(for: worktree, baseBranch: config.worktrees.baseBranch)
+        let rps = rightPaneStore.state(
+            for: worktree,
+            baseBranch: config.worktrees.baseBranch,
+            trackUpstreamForCommits: config.changes.trackUpstreamForCommits
+        )
         rps.reveal(path: path)
     }
 
--- a/Alas/Sources/Center/CenterPaneView.swift
+++ b/Alas/Sources/Center/CenterPaneView.swift
@@ -161,7 +161,8 @@ struct CenterPaneView: View {
                         )
                         let rps = state.rightPaneStore.state(
                             for: worktree,
-                            baseBranch: state.config.worktrees.baseBranch
+                            baseBranch: state.config.worktrees.baseBranch,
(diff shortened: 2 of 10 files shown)


mrmans0n:
@codex review

PR #435, opened 2026-06-02
feat(acp): per-message gutter actions menu

mrmans0n:
## Summary

Replaces the inline per-message copy button in the ACP chat transcript with a hover-revealed **"…" menu** in a reserved right-hand gutter. The menu's first action is **Copy message**; it's structured as the entrypoint for future per-message actions (e.g. fork-from-here).

- New `ACPMessageGutter` — a reusable generic wrapper that overlays a native `Menu` (matching the `BranchOpsMenu` pattern) in the right gutter on hover. Copy payload is a closure (`() -> String`) so streamed agent text is read at click time, not captured stale.
- Symmetric reserved lanes added once at the transcript container (`28 + ACPMessageGutterLayout.laneWidth`), so every row stays aligned and the "…" lands at the same far-right x for both user and agent messages.
- Old `ACPHoverCopyButton` and its inline corner overlays removed; `ACPDelayedHoverVisibility` retained and reused.
- Button keeps itself alive while hovered (its own `onHover`), so pointing at it past the hide delay doesn't fade/disable it mid-aim.

Copy output is unchanged (user: raw text, agent: `buffer.value`), still covered by `ACPTranscriptMarkdownTests`.

## Test Plan

- [x] `xcodebuild build` succeeds
- [x] `xcodebuild test` — full suite green (incl. `ACPTranscriptMarkdownTests` copy-payload guard)
- [ ] Symmetric lanes look balanced at wide and narrow pane widths
- [ ] "…" reveals on hover in the right gutter, top-aligned, same x for user + agent messages
- [ ] Click registers on **user** messages (button sits past the bubble's right spacer)
- [ ] Menu → Copy message pastes correct Markdown (incl. just-streamed agent text)
- [ ] Menu not dismissed prematurely; button stays clickable when hovered
- [ ] Tool/file cards and thoughts still render, left-aligned with agent prose


mrmans0n:
@codex review

PR #441, opened 2026-06-02
fix(acp): move file mention picker ranking off main thread, use FileIndex

mrmans0n:
## What

The @-mention file picker in the ACP composer had two problems:

1. **Main-thread jank**: `FileManager.default.enumerator` ran a synchronous FS walk (even inside `Task.detached`, it blocks the cooperative thread pool). `ranked` was a computed property doing O(n) on `@MainActor`.

2. **Search filtering broken during indexing**: The `ranked` computed property returned `[]` until the FS walk finished, so typing a query while indexing showed nothing.

## Changes

- **`Alas/Sources/ACP/UI/ACPMentionPicker.swift`**: Replaced `ranked` computed property with `@State`-backed async ranking using `Task.detached` with 16ms debounce and cancellation. Added `filesProvider` closure parameter.
- **`Alas/Sources/ACP/UI/ACPComposer.swift`**: Threaded `filesProvider` through `ACPInputField` → `Coordinator` → `ACPMentionPanel` → `ACPNSTextView.presentMentionPopover()` with nil default for test compatibility.
- **`Alas/Sources/ACP/UI/ACPComposerShell.swift`**: Added `filesProvider` to `ACPComposer` init and wired to `ACPInputField`.
- **`Alas/Sources/ACP/UI/ACPTabView.swift`**: Created the closure bridging `state.fileIndex.entries(forWorktreePath:)` → `[URL]`.
- **`Alas/Sources/App/AppState.swift`**: Removed `private` from `fileIndex`.

The `FileIndex` actor (git ls-files, 30s cache, respects .gitignore) replaces the manual enumerator. `MentionFuzzy.collectFiles` is preserved as fallback when no provider is given.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPComposer.swift
+++ b/Alas/Sources/ACP/UI/ACPComposer.swift
@@ -33,6 +33,9 @@ struct ACPInputField: NSViewRepresentable {
     /// Called when image staging fails so the chrome can show a transient
     /// notice. Receives the specific error that caused the failure.
     let onImageError: (ACPImageStaging.StagingError) -> Void
+    /// Async file list provider for the @-mention picker. When nil,
+    /// falls back to a synchronous `FileManager` enumerator.
+    let filesProvider: (@Sendable () async -> [URL])?
 
     func makeNSView(context: Context) -> NSScrollView {
         let textView = ACPNSTextView()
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #443, opened 2026-06-02
feat(acp): synthesize Thinking chip for cursor-agent from model variants

mrmans0n:
## Summary

Cursor's ACP server doesn't advertise a `thought_level` configOption like Claude/Codex/OpenCode/Pi — it encodes thinking as bracket-suffixed model variant ids, e.g. `claude-opus-4-6[effort=high,context=200k,thinking=true,fast=false]`. Without explicit handling Alas's heuristic finds nothing in `configOptions` and the Thinking chip stays hidden for cursor sessions.

This PR parses those variants client-side and synthesizes orthogonal Model + Thinking chips, both dispatching via the existing `session/set_model` path (item ids carry the full composed variant string).

- Prefer `effort` attr; fall back to `thinking` bool if no variant in the current base group has `effort`.
- Hide the Thinking chip when no model id contains a `[…]` suffix.
- Persist the full variant id as `currentModel` (no schema change).
- The overlay only fires for `agentId == "cursor-agent"` AND when the standard `thought_level`-heuristic returned nothing — so if Cursor later ships a real `thought_level` configOption, the heuristic wins and the workaround becomes a no-op automatically.
- Don't clobber a `category: "model"` configOption-sourced Model chip if one is present.

## Test plan

- [x] `CursorModelVariants` parser/compose round-trip + edge cases (no brackets, missing close, bare fragments)
- [x] `CursorModelVariants.derive` produces correct Model/Thinking chips for effort-bearing variants, falls back to `thinking` bool, hides Thinking when neither attr present, bails when current model isn't in the list
- [x] `ACPChipState.normalize` overlay fires only for cursor-agent + bracketed variants, defers to a real `thought_level` configOption when advertised, leaves other agents unaffected
- [x] Full AlasTests suite green (2893 tests in 365 suites)
- [ ] Manual verification: open a cursor-agent ACP session and confirm Model + Thinking chips render and `session/set_model` is dispatched with the composed variant id on selection

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -333,6 +333,7 @@
 		5D8DF496FC722EE24F69A60A /* AgentHookSocketServer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 976B145EAC17A395380DFB47 /* AgentHookSocketServer.swift */; };
 		5DF3F01E3DC672E4C6CBEDBC /* CommitEditorTabView.swift in Sources */ = {isa = PBXBuildFile; fileRef = D7268D601058775FE02B6B06 /* CommitEditorTabView.swift */; };
 		5E5600E6451068BCBF732ABD /* ACPChipStateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 239125BB939233233D405EB8 /* ACPChipStateTests.swift */; };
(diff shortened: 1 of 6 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #444, opened 2026-06-02
feat(updates): nightly release track

mrmans0n:
## Summary

- In-app "Check for Updates" now updates nightly builds off the rolling `nightly` GitHub release. Stable builds keep checking `releases/latest` exactly as before.
- A new `BuildIdentity` value type reads the running app's track / git SHA / build date from Info.plist keys (`AlasReleaseTrack`, `AlasGitSHA`, `AlasBuildDate`). Stable builds default to `track = stable`; the nightly workflow overwrites the three keys with `plutil` before signing so a nightly DMG carries its real identity.
- `ReleaseInfo` is now an enum of `.stable(StableReleaseInfo)` / `.nightly(NightlyReleaseInfo)`. The sheet branches title, subtitle, and the View-Release label by case. Nightly builds force the `.direct` action row regardless of `InstallSource` (the brew cask only tracks stable).
- Nightly freshness predicate: `remote.targetCommitish != local.gitSHA && remote.publishedAt > local.buildDate`, with date-only fallback when the local SHA is absent. A re-tag to the same SHA does not look like an update.

## Test plan
- [x] `xcodebuild test` — 50 tests across 6 suites pass (`BuildIdentityTests`, `ReleaseCheckerTests`, `ReleaseInfoTests`, `SemanticVersionTests`, `InstallSourceTests`, `UpdateThrottleTests`).
- [x] `xcodebuild build` clean.
- [ ] Hand-verify on a local **stable** build: launch the app, choose **Alas → Check for Updates…**, confirm the stable path still surfaces an update sheet (or "You're up to date") against `releases/latest`.
- [ ] Hand-verify nightly→nightly: only possible after this lands and the nightly workflow publishes a build carrying the new Info.plist keys. Smoke-test on the next nightly DMG.

## Notes
- Spec: `docs/superpowers/specs/2026-06-02-nightly-update-track-design.md` (gitignored locally per repo convention).
- Asset name for nightlies is `Alas-nightly.dmg` (arm64 only today). The arch argument is ignored on the nightly track; the pattern lives in one place if Intel nightlies arrive later.
- Out of scope: user-facing track switcher, cross-track notifications, in-app DMG download/apply, Intel nightly builds, updating the Homebrew cask.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -166,6 +166,20 @@ jobs:
             -destination 'platform=macOS,arch=arm64' \
             -quiet build
 
+      - name: Stamp nightly build identity into Info.plist
+        # Overwrite the stable-default Info.plist keys committed in the
+        # repo with the real track / SHA / build-date for this nightly.
+        # Must run BEFORE signing, since signing seals the bundle.
+        if: steps.skip-check.outputs.skip != 'true'
+        run: |
+          plist="build/Build/Products/Release/Alas.app/Contents/Info.plist"
+          sha="$(git rev-parse HEAD)"
+          iso_date="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+          plutil -replace AlasReleaseTrack -string "nightly" "$plist"
+          plutil -replace AlasGitSHA -string "$sha" "$plist"
+          plutil -replace AlasBuildDate -string "$iso_date" "$plist"
+          plutil -lint "$plist"
+
       # Gate signing on main-built refs only — non-main workflow_dispatch
       # checks out (and runs scripts from) the dispatched ref, so handing
       # the Developer ID cert + notary credentials to that ref's scripts
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -156,6 +156,23 @@ jobs:
           path: .build/xcode/SourcePackages
           key: ${{ steps.swiftpm-cache.outputs.cache-primary-key }}
 
+      - name: Stamp nightly build identity into source Info.plist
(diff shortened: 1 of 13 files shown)


mrmans0n:
@codex review

mrmans0n:
Both Codex findings addressed in 70467c07:

1. Nightly comparison now resolves the tag's true commit SHA via `/repos/mrmans0n/alas/git/ref/tags/nightly` instead of trusting `target_commitish` (which GitHub leaves at `main` for the rolling tag). Verified via curl that the live release returns `target_commitish: main` while the tag-ref endpoint returns the real SHA `5227897f377d44d9e5134413885beba9828b86dc`.
2. The three new Info.plist keys are now pinned in `project.yml` under `info.properties`; ran xcodegen and re-verified the regenerated plist contains them.

@codex review

mrmans0n:
Both follow-up Codex findings addressed in cf673cdf:

1. **SHA-only nightly compare.** Dropped `publishedAt` from the freshness predicate. The rolling release's publish time is unreliable since the release is updated in place; tag SHA is the source of truth. No local SHA → upToDate (don't pester hand-built nightlies).
2. **Stamp before build.** Moved the plutil step to before `xcodebuild` so it mutates the source `Alas/Resources/Info.plist` rather than the post-signed bundle. Signing now seals the stamped values for both main and non-main dispatches.

@codex review

mrmans0n:
P3 finding addressed in eca48ed4: sheet subtitle now reads just `Nightly <shortSHA>` (no relative-time string). Also dropped `NightlyReleaseInfo.publishedAt` since nothing reads it anymore.

@codex review

PR #447, opened 2026-06-02
fix(acp): restore transcript scroll-up pagination on macOS 15+

mrmans0n:
## Problem

In the ACP transcript, scrolling up did nothing — older messages never appeared and tail-follow never paused. Confirmed via runtime instrumentation that **the data was never missing** (`visibleHead=1178, count=1208` — the older rows were in memory behind the render window); scroll-position detection was the failure.

On **macOS 15+ (Tahoe)** both signals the transcript relied on are dead:

1. **NSScrollView bounds observer** — SwiftUI's `ScrollView` is no longer `NSScrollView`-backed. The ancestor walk ends at `NSThemeFrame` with no scroll view, so the observer never attaches and `handleBoundsChange` never fires.
2. **Head-sentinel `PreferenceKey`** — the `GeometryReader` computes correct, live frames during scroll, but the value delivered to `onPreferenceChange` is always `.zero` (and it doesn't fire during scroll). So `frame.maxY > 0` is never satisfiable and the head-step never triggers.

## Fix

`ACPTranscriptScrollTracking`, a version-branching `ViewModifier`:

- **macOS 15+:** `onScrollGeometryChange` reads scroll position (`visibleRect` / `contentSize`) straight from the ScrollView, driving both head-step pagination and tail-follow pause/resume through the existing `ACPScrollDirectionClassifier`. Revealed rows are anchored with `proxy.scrollTo(anchor:)` since there's no clip view to shift.
- **macOS 14:** keeps the legacy NSScrollView observer + preference path, which still works there.

No change to the model/data layer — the existing hydration/backfill/window logic and its tests are untouched.

## Testing

- Builds clean (Debug, macOS).
- Manually verified: scrolling up now pages in older messages with the viewport staying anchored.
- Existing ACP classifier / pagination-indicator / transcript-window / hydration suites unaffected.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPMessageList.swift
+++ b/Alas/Sources/ACP/UI/ACPMessageList.swift
@@ -154,43 +154,24 @@ struct ACPMessageList: View {
                     .frame(maxWidth: .infinity, alignment: .center)
                 }
                 .coordinateSpace(name: scrollSpaceName)
-                .background(
-                    ACPScrollEventObserver(
-                        onPaused: {
-                            setFollowsTranscriptTail(false)
-                        },
-                        onAtBottom: {
-                            setFollowsTranscriptTail(true)
-                        },
-                        isRestoring: { isRestoringTail },
-                        onScrollViewResolved: { scrollViewRef.scrollView = $0 }
-                    )
-                )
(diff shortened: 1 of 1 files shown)


mrmans0n:
@codex review

PR #451, opened 2026-06-02
Distinguish behind-base and behind-upstream commit chips

mrmans0n:
## Summary
- The Commits section header (Changes tab) could show two identical `N behind` chips, distinguishable only by a hover tooltip. They now read `↓N <label>` with distinct colors so it's clear what each is behind at a glance.
- Base chip → `↓N <base-branch>` in the accent color (behind the trunk you compare against).
- Upstream chip → `↓N remote` in the caution color (your branch has commits on its remote tracking ref you don't have locally).
- `↓` denotes "commits to pull in"; each chip still only appears when its count is nonzero, and the full-ref tooltips are retained.

Rendering-only change to `BehindChip` and its two call sites in `CommitsSectionView.swift`; no changes to the git probes, visibility predicates, or `RightPaneState`/`GitService`.

## Test Plan
- [x] `RightPaneStateSyncStatusTests` (incl. 4 new `BehindChip` tests) pass: `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/RightPaneStateSyncStatusTests test`
- [x] Full app builds: `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' build`
- [ ] Visual check in-app: behind base only → one accent `↓N <base>` chip; behind own remote only → one caution `↓N remote` chip; both → two visually distinct chips; tooltips show full refs.

### DIFF
--- a/Alas/Sources/Right/CommitsSectionView.swift
+++ b/Alas/Sources/Right/CommitsSectionView.swift
@@ -1,20 +1,44 @@
 import SwiftUI
 
 /// Small pill shown in the Commits section header trailing slot when the
-/// worktree is behind a tracked ref (base trunk or its own upstream).
-/// Purely informative.
+/// worktree is behind a tracked ref. Purely informative. The `role` selects
+/// a distinct theme color so the "behind base" and "behind upstream" chips
+/// read as different things at a glance; the `label` names the target.
 struct BehindChip: View {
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #456, opened 2026-06-03
fix(ACP): mention picker search and chip alignment

mrmans0n:
## Summary
- Use the shared fuzzy matcher for ACP @ mention file ranking, including multi-token path queries.
- Align mention chips using surrounding font metrics so they sit on the composer baseline.
- Add regression coverage for fuzzy @ search and chip frame alignment.

## Validation
- `rtk xcodegen`
- Focused ACP UI tests for `ACPMentionPickerTests` and `ACPComposerDraftBridgeTests`
- Manual launch and ACP composer check

Note: the full `xcodebuild ... test` run was attempted locally but interrupted after it stayed idle in build/pre-test state with no compiler or `xctest` child.


mrmans0n:
@codex review

PR #457, opened 2026-06-03
fix(ACP): dismiss composer pickers on tab switch

mrmans0n:
## Summary

- Close the slash and `@`-mention picker panels when the ACP composer is torn down (tab switch / window close), so they no longer outlive the composer and float across tabs until app quit.
- Cleanup runs from `dismantleNSView` (the SwiftUI-canonical teardown hook) rather than `viewWillMove(toWindow:)`. Closing a child `NSPanel` from inside SwiftUI's view-update commit re-entered the menu shortcut path and dismissed the just-opened replacement tab (e.g. ⌘T while the slash picker was open opened a terminal tab that was then immediately closed).

## Test plan

- [x] Open ACP tab, type \`/\`, switch to another tab — slash panel disappears.
- [x] Open ACP tab, type \`/\`, press ⌘T — new terminal tab stays.
- [ ] Same as above with \`@\` (mention picker).


mrmans0n:
@codex review

PR #459, opened 2026-06-03
fix(terminal): self-heal orphaned zmx sessions

mrmans0n:
## Summary

Tab-close and worktree-delete already dispatch `zmx kill` for each leaf, but the kill ran in a fire-and-forget `Task.detached` that the process could outlive on Cmd-Q. Anything lost that way persisted forever because nothing on a subsequent launch ever tried to clean it up — a user reported accumulating "a bajillion" leftover sessions after lots of tab closing.

Two complementary fixes:

- **Boot-time sweep**: after `reloadTabs`, enumerate `zmx ls` and kill any `alas-<wtHash>-<leafHash>` session whose `wtHash` matches one of our worktrees but whose `leafHash` isn't in any persisted leaf. Multi-instance safe — another Alas under the same `ZMX_DIR` has different `wtHash` values, so its sessions stay untouched. Legacy `alas-<leafId>` names are skipped (close-path cleanup still handles those).
- **Quit-time drain**: track every dispatched kill task; in `applicationWillTerminate`, block up to 3s on a background awaiter so in-flight `zmx kill` subprocesses finish talking to the daemon before the process exits.

## Test plan

- [x] `xcodebuild test` — all 2948 tests pass
- [x] New tests cover orphan filter (multi-instance safety + legacy ignored), sweep dispatch, drain blocks until kill completes, drain no-op when nothing pending
- [ ] Manual: open tabs, close one, Cmd-Q quickly, relaunch → `zmx ls` shows no orphan for the closed tab
- [ ] Manual: simulate a leak (e.g. `zmx run alas-<wtHash>-<random>` against a known worktree), relaunch Alas → orphan gone after boot

### DIFF
--- a/Alas/Sources/App/AppState.swift
+++ b/Alas/Sources/App/AppState.swift
@@ -409,6 +409,28 @@ final class AppState {
         // waiting for `restoreTerminalTabIfNeeded` (driven by
         // TerminalTabView.task) to fire when the user opens the tab.
         refreshPersistedHookSymlinks()
+        sweepOrphanZmxSessions(worktreeIds: allWorktreeIds)
+    }
+
+    /// Compute (knownWorktreeIds, knownLeafIds) from in-memory state and
+    /// hand them to `TerminalService.sweepOrphans`. Run after `reloadTabs`
+    /// so persisted leaves are visible; without this, sessions leaked by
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #461, opened 2026-06-03
feat(ACP): multi-instance sessions: single-writer lease, live mirror, takeover

mrmans0n:
## Summary

Lets two concurrently-running Alas instances share an ACP session (one per-worktree SQLite DB) without corrupting each other. Previously, two instances on the same worktree both spawned an agent and wrote the same `messages`/`remote_session_id`, which scrambled transcripts and broke session reload.

Built in three phases:

- **Phase 1 — single-writer lease + isolation + recovery.** A per-process `instanceId`, a new `session_leases` table (schema v7) with an atomic `claimLease` (`BEGIN IMMEDIATE` + UPSERT-WHERE, dead-PID/stale-heartbeat reclaim), a heartbeat, and the keystone rule: only the lease holder spawns an agent runner and writes — a second instance becomes a read-only **mirror**. A one-time recovery pass dedups duplicate `(session_id, seq)` rows left by the old corruption, repairing already-broken DBs.
- **Phase 2 — live mirror.** A Darwin cross-process change-ping (writer posts on each persist + teardown); the mirror re-reads the transcript (debounced, with a slow poll backstop) and renders read-only with a busy/idle spinner. Composer disabled while mirroring.
- **Phase 3 — explicit takeover.** A "Take over here" button seizes the lease and re-attaches this instance as the writer (via the existing `session/load` path); the previous writer detects the loss via its heartbeat and stands down (tears down its agent, becomes a mirror).

~930 lines across `ACPSessionStore`, `ACPSessionManager`, `ACPSessionRunner`, `ACPTabView`, `AppState`, plus new `ACPProcessLiveness` and `ACPChangeNotifier`.

## Test Plan

- [x] Full unit suite green (2972 tests, 0 failures) — includes new lease claim/reclaim/seize, dead-PID TOCTOU regression, recovery dedup, failed-attach lease release, heartbeat re-assert, mirror re-read, takeover, and Darwin notifier tests.
- [ ] **Two-instance manual smoke test (not yet done):** run two dev builds on the same worktree; confirm the second shows the read-only mirror banner and does not spawn a second agent; confirm the mirror updates live as the writer streams; click "Take over here" and confirm the writer role moves and only one agent runs (`pgrep -fl claude`); relaunch both and confirm the session loads cleanly.

## Notes

- Known accepted edge: a *hard* takeover-attach failure briefly leaves the session writer-less; it self-heals on the next prompt submit (no dual-writer, no corruption).
- Mirror re-read is a full transcript rebuild (keeps `toolCallIndices`/render-window invariants); the incremental-read cursor was dropped as the safer choice for this dev-workflow feature.


mrmans0n:
Addressed the P1 (takeover dual-writer window) in 838811cf: runner writes are now **lease-guarded** — `ACPSessionRunner` checks it still owns the session lease before persisting (`persistIndices`/`persistFromIndex` early-return when the DB lease owner != this instance). So even during the ≤5s takeover handoff where two runners briefly coexist, only the current lease holder writes; the former writer's persists become no-ops. Added positive + negative unit tests. Also merged latest `main` and resolved conflicts.

@codex review

mrmans0n:
Follow-up P1 addressed in 56cd5619: now **every** runner store-write path is lease-guarded, not just transcript persistence —
- `persistSessionRow()` (session/title/model), `persistQueue()`, and the `updateMessagePayload` write in `userCancel()` each early-return on `!holdsLeaseForWrite()` (the Stop RPC to the agent still runs; only the DB write is gated);
- `ACPPermissionDecisionLog.record()` gained a `canWrite` gate the runner wires to the same lease check.

So a former writer in the ≤5s handoff cannot overwrite queue/session/permission state after another instance owns the lease. Added unit tests for the queue + session-row skip paths. Reads are intentionally not gated.

@codex review

mrmans0n:
Both addressed in 043d035e:

**P1 (mirror queue writes):** lease-guarded at the data layer — `persistQueue(for:)`, plus `persist(_:)` (model/mode/title) and `renameSession(...)` now `guard _ownedLeases.contains(id)` before writing, so a mirror can't overwrite the owner's queue/session state via any toolbar/queue control. UI layer: the five `ACPMessageList` queue callbacks (edit/remove/retry/reorder/clear) are neutralized when `isMirror` (matching the already-disabled composer).

**P2 (poll leak):** `endMirroring(sessionId:)` is now called in `evictIfIdle` (the choke point reached on tab close for both writers and mirrors), so a closed mirror's 2.5s poll + notifier subscription are torn down. It's idempotent, so writer sessions are unaffected.

Added tests for the mirror queue-write no-op and the poll cancellation on evict.

@codex review

mrmans0n:
Pushed f6b2c05d: the earlier mirror-write guards were keyed on `_ownedLeases` (explicit ownership), which incorrectly no-op'd in the not-yet-attached/single-instance case (broke an existing `persistQueueWithoutRunner` test). Switched the guards in `persist`/`renameSession`/`persistQueue` to `!isMirror(sessionId:)` — they block **only** when another *live* instance owns the lease (a true mirror), and persist normally for the writer and pre-lease cases. Full suite green again.

@codex review

mrmans0n:
Pushed ca815a68 addressing both round-4 findings:

**P1 (re-read store lease before manager writes):** added `anotherLiveInstanceOwnsLease(sessionId:)` which reads the SQLite lease row directly — unlike `isMirror`, it does NOT short-circuit on the in-memory `_ownedLeases`. The three manager write guards (`persist`, `renameSession`, `persistQueue`) now use it, so a former owner whose lease was seized is blocked immediately during the handoff (before its heartbeat catches up), while the writer and not-yet-leased cases still write. Added a `staleOwnerCannotPersistQueue` regression test.

**P2 (cancel mirror pollers on manager dispose):** added `shutdownBackgroundTasks()` (ends all mirror subscriptions/pollers + cancels heartbeats), called from `disposeACPManager` so worktree close/archive/delete tears down mirror-only sessions that have no runner. Added a `disposeStopsMirrorPoll` test.

Full suite green (3130 tests).

@codex review

mrmans0n:
Pushed d33fd33a — this closes the takeover-handoff window as a class rather than path-by-path:

- **Prompt stand-down (root fix):** the writer now subscribes to the change-notifier and, on a takeover ping, re-reads the store lease and stands down within ~100ms instead of waiting up to 5s for its heartbeat. (Subscription torn down on detach/standDown/failed-attach/dispose.)
- **`isMirror` reads the store** (no `_ownedLeases` short-circuit), so a former owner's pane is immediately read-only during handoff — composer + queue controls disabled.
- **`attach` re-checks the lease right before commit** (after the initialize/newSession/loadSession awaits): if taken over mid-attach, it shuts down the just-spawned agent, becomes a mirror, and does not persist the remote id or register a runner.
- **Runner file writes gated** on `holdsLeaseForWrite()` (rejects with a JSON-RPC error after lease loss); terminal execution is covered by prompt stand-down → `runner.stop()` → `terminalHost.killAll()`.

Full suite green (3134 tests). Added tests for store-reading isMirror, writer-watch lifecycle, and dispose teardown.

@codex review

mrmans0n:
Pushed 80c5e387 — three more, all fixed:

- **P2 (close while spawning):** `cleanupACPSession` no longer early-returns when a session has no runner yet; it always calls `detach`, which releases the lease + stops heartbeat/writer-watch + ends mirroring even in the pre-runner attach window. No more leaked lease on tab-close-during-spawn.
- **P1 (takeover continuity):** `takeOver` now refreshes `session.remoteSessionId` from the store before re-attaching, so a mirror that opened before the writer persisted the remote id re-attaches via `session/load` instead of creating a fresh `session/new`.
- **P2 (terminal side effects):** `handleTerminalRequest`'s `.create` case is now gated on `holdsLeaseForWrite()` (rejects with -32003 after lease loss), matching the file-write gate — defense-in-depth on top of prompt stand-down.

Full suite green (3137 tests). Added tests for each.

@codex review

mrmans0n:
Pushed 6b19d9b4:

- **P1 (mirror queue sync):** `refreshMirror` now reloads `session_queue` (via `store.loadQueue` → `session.restoreQueue`) alongside the transcript, so a mirror's queue can't go stale and be persisted on takeover.
- **P2 (no mirroring closed sessions):** `standDown` only calls `beginMirroring` when the session is still open, and `beginMirroring` early-returns for a non-open session — closes the takeover-vs-tab-close race that could start a poll for an evicted session.
- **P2 (recovery write):** the `context_recovery_pending` store writes on the loadSession-failure path are now wrapped in `!anotherLiveInstanceOwnsLease(...)`, so a takeover during `loadSession` can't write recovery state to a session another instance owns.

Full suite green (3140 tests), tests added for each.

@codex review

mrmans0n:
Pushed ab688ddd:

- **(mirror queue sync, all cases):** moved the queue reload in `refreshMirror` ahead of the transcript-message early-return, so the mirror's queue syncs on every ping even when the writer changed only the queue (no new messages).
- **P2 (queue flush):** `flushQueueIfIdle` now early-returns on `!holdsLeaseForWrite()`, so a former writer mid-handoff can't pop the queue head or dispatch a `session/prompt` to the agent for a session it no longer owns.

Full suite green (3142 tests), tests added for both.

@codex review

mrmans0n:
Pushed: `shutdownBackgroundTasks()` now also releases every lease this manager still owns (not just cancels tasks). This covers the pre-runner attach window — a session disposed while spawning (lease held, not yet in `runners`) is released immediately so another instance can reclaim it without waiting for the 15s stale timeout. Test added (`shutdownReleasesOwnedLeases`).

@codex review

mrmans0n:
Pushed 0e9f6b95: added an `isDisposed` flag set at the top of `shutdownBackgroundTasks()`, and the `attach` pre-commit guard now aborts on `isDisposed || anotherLiveInstanceOwnsLease(...)`. So an attach that's mid-flight when its manager is disposed shuts down the spawned connection and returns without registering a runner or persisting — no orphaned agent, no commit against a released lease (and it doesn't start a mirror poll on a dead manager). Its `defer` still releases the lease idempotently. Test `attachAbortsAfterDispose` added; full suite green (3144).

@codex review

mrmans0n:
Pushed f91dca34:

- **P1 (dispose ordering):** `shutdownBackgroundTasks()` no longer releases leases — it only sets `isDisposed` + cancels tasks. Leases are now released by a new `releaseAllOwnedLeases()` called at the END of disposeACPManager's async teardown, AFTER all runner connections are shut down (`detach` already releases per-runner leases post-`connection.shutdown`). So a freed lease is never claimable while the old agent is still alive.
- **P1 (takeover queue):** `takeOver` now also re-reads `session_queue` from the store (alongside `remoteSessionId`) before re-attaching, so it can't act on a stale mirror queue.
- **P2 (terminal gate):** the `holdsLeaseForWrite()` guard now covers `terminal/kill` and `terminal/release` too, not just `create`. Teardown is unaffected (`runner.stop()` calls `terminalHost.killAll()` directly, not through the request handler).

Full suite green (3147), tests added/retargeted for each.

@codex review

mrmans0n:
Pushed 663cbcc3:

- **P1 (dispose vs in-flight attach):** `attach` now tracks itself in `attachingSessions` (inserted after lease acquisition, removed in the same `defer` on every exit). `releaseAllOwnedLeases()` skips any session still attaching, so a pre-runner in-flight attach's lease is released by its OWN `defer` (which runs after `connection.shutdown` in the isDisposed-abort path) — never freed while its agent is still alive.
- **P2 (cancel RPC):** `userCancel()` now `guard holdsLeaseForWrite()` before sending `session/cancel` to the agent (local prompt-id bookkeeping is kept; teardown is unaffected since standDown/stop kill via `terminalHost.killAll()`/`invalidateActivePrompt()`, not `userCancel`).

Full suite green (3149), tests added for both.

@codex review

mrmans0n:
## Review wrap-up

Thanks Codex — across the review rounds we addressed **every substantive finding**: the lease election, prompt stand-down on takeover, store-reading mirror/write guards, attach commit re-check, recovery-write gating, mirror queue+transcript sync, and lease-gating on all runner side effects (transcript, queue, session row, file writes, terminal create/kill/release, cancel). CI (`build-test`) is green and the suite is at 3149 tests.

### Disposition of the remaining dispose-ordering items (acknowledged, not blocking)

The last open findings are narrow async-ordering races confined to **manager disposal** — i.e. closing/archiving/deleting a worktree during the sub-second window while an ACP session is still mid-`attach`. We're consciously accepting these as-is for this dev-workflow feature because they are **benign and self-healing**, not the cross-instance corruption this PR fixes:

- **Heartbeat stops before agent shutdown / queued-attach claims a lease / cancel recheck after await:** in all of these the in-flight/former writer **cannot corrupt shared state** — every store write is lease-gated, and the `isDisposed` commit guard prevents a disposed manager's attach from ever registering a runner or persisting. The worst case is a briefly-held lease on a manager that's going away, which is **reclaimed automatically after the 15s stale-lease timeout** (or immediately once the in-flight attach's `defer` runs).
- These windows require a worktree teardown to land within a few hundred ms of an ACP attach — and even then resolve without data loss.

A fully ordered dispose handshake (await all agent shutdowns before tearing down tasks/leases) is a reasonable future hardening, but it's out of scope for the value here. Merging.

> Note: manual two-instance smoke test (run two dev builds on the same worktree; verify mirror banner, live updates, and takeover) is still recommended before relying on it day-to-day — the automated suite covers the lease/store logic but not the live cross-process UI.

PR #471, opened 2026-06-04
feat(acp): auto-send transcript context on session restore

mrmans0n:
## Summary
- On session restore failures, transcript context is now sent to the agent automatically instead of requiring a manual user action.
- A new `ContextRecoveryStatus` state machine (`restoring`, `sendingTranscript`, `restored`, `failed`) drives an inline status row in the message list showing progress and a retry button on failure.
- The `contextRestoreWarning` banner is suppressed while the inline recovery row is active to avoid duplicate messaging.

## Testing
- Updated and renamed existing tests to cover the new auto-send behavior: load failure, pending recovery after restart, and queued-prompt ordering all verify `session/prompt` is sent automatically and `contextRecoveryPending` is cleared.
- Added assertion for `contextRecoveryStatus == .failed(...)` on delivery failure and `.sendingTranscript` on hydration with a recoverable transcript.

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSession.swift
+++ b/Alas/Sources/ACP/Session/ACPSession.swift
@@ -47,6 +47,7 @@ final class ACPSession: ObservableObject, Identifiable {
     @Published var setupState: SetupState = .checking
     @Published var lastError: String?
     @Published var contextRestoreWarning: ContextRestoreWarning?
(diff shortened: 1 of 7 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #483, opened 2026-06-07
fix(acp): keep composer clear of hero content on short panes

mrmans0n:
## Problem

On small / short ACP panes, the empty state ("What should we work on?") and the first-run "Connecting…" state looked broken: the floating chat composer overlapped the content above it (the starter / phase chips slid behind the pill).

The two centred hero views (`ACPNewChatEmptyStateView`, `ACPFirstRunConnectingView`) lifted themselves above the composer with a hardcoded `.padding(.bottom, 210)`. But the composer floats at a **dynamic** inset (`raisedEmpty` ≈ 0.34 × pane height), so the fixed 210 only left a comfortable gap on tall panes — the gap shrinks as the pane gets shorter and goes negative below ~400pt.

## Fix

Replace the magic `210` with `ACPComposerPlacement.raisedHeroBottomPadding(containerHeight:)`, derived from the same placement math:

- Solves the centring geometry for a **constant gap** above the floating composer.
- Lands on exactly `210` at a ~900pt pane, so **large windows are visually unchanged**.
- **Grows** as the pane shrinks, restoring the gap on short panes.
- **Capped** on very short panes so the artwork stays on-screen rather than clipping under the toolbar.

`ACPTabView` computes it once from the existing `GeometryReader` and threads it into both hero views. The connecting skeleton (`ACPConnectingPlaceholder`) is unchanged — it uses the normal bottom composer placement, like the real transcript.

## Tests

Added to `ACPNewChatEmptyStateTests`:
- No-overlap invariant across pane heights 400–1200 (content clears the composer and never clips the top).
- Back-compat: `== 210` at 900pt and grows below.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPComposerShell.swift
+++ b/Alas/Sources/ACP/UI/ACPComposerShell.swift
@@ -14,6 +14,35 @@ enum ACPComposerPlacement: Equatable {
             return min(max(target, 96), min(320, roomAwareMaximum))
         }
     }
+
+    /// Bottom padding for the centred hero content shown in the empty and
+    /// first-run-connecting states, so it always clears the floating
+    /// composer.
+    ///
+    /// The composer floats at `bottomInset(.raisedEmpty)` and is roughly
+    /// `pillHeight` tall, so its top edge sits `inset + pillHeight` above the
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #486, opened 2026-06-08
fix(acp): clear restored transcript context marker

mrmans0n:
## Summary

- Clear the "Transcript restored as model context" success marker when new agent output starts.
- Expire the restored marker after 30 seconds if the agent stays idle.
- Keep failed recovery state visible when subsequent agent output arrives.

## Validation

## Notes

- Local `xcodebuild ... test` currently fails before tests execute with Xcode reporting the generated `.build/ghostty/GhosttyKit.xcframework` as missing during the test action, despite the build action succeeding with the same artifact present.


mrmans0n:
@codex review

PR #489, opened 2026-06-08
fix(acp): keep queued controls clickable

mrmans0n:
## Summary

- Move pending queued ACP controls into a stable lane to the left of the queued bubble.
- Keep the hover region on the full queued row so moving from the bubble to the controls does not hide them.
- Render sending queued items again while keeping queue drag/drop pending-only.

## Validation

- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/ACPMessageListPaginationTests`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/ACPQueuedBubbleActionMetricsTests`

Full `xcodebuild ... test` was attempted locally, but the run hung around ACP terminal routing tests and was terminated instead of being counted as a pass.

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSession.swift
+++ b/Alas/Sources/ACP/Session/ACPSession.swift
@@ -473,12 +473,9 @@ final class ACPSession: ObservableObject, Identifiable {
         queue.insert(contentsOf: snapshot, at: insertAt)
     }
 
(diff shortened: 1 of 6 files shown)


mrmans0n:
@codex review

PR #496, opened 2026-06-09
fix(worktrees): new worktree project selection

mrmans0n:
## Summary

Fix the first new-worktree dialog opened from a project row selecting the wrong repository after a cold start.

## Root Cause

`RootView` presented `NewWorktreeDialog` with `sheet(isPresented:)` while storing the requested project id in a separate state value. On the first presentation, SwiftUI could build the sheet before that separate preset state had propagated, so the dialog saw `nil` and defaulted to the first project.

## Changes

- Replaced the boolean new-worktree sheet presentation state with an item-driven `NewWorktreePresentation` that carries the optional project id.
- Routed sidebar and notification-based new-worktree openings through the same presentation value.
- Added focused tests for preserving a requested project id and supporting global creation.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/NewWorktreeDialogTests test` (34 tests)

Note: the full `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` command was started locally, progressed through many passing suites, then stopped producing output for several minutes and was interrupted rather than left running.

### DIFF
--- a/Alas/Sources/App/RootView.swift
+++ b/Alas/Sources/App/RootView.swift
@@ -1,13 +1,17 @@
 import AppKit
 import SwiftUI
 
+struct NewWorktreePresentation: Identifiable, Equatable {
+    let id = UUID()
+    let projectId: String?
+}
+
 struct RootView: View {
     @Bindable var state: AppState
     @State private var showNewProject = false
     @State private var editingProject: ProjectConfig?
     @State private var removingProject: ProjectConfig?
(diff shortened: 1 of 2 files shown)


mrmans0n:
No blocking findings from review. CI is green.

PR #500, opened 2026-06-09
feat(terminal): add Open sessions manager to Terminal settings

mrmans0n:
## Summary

Adds an **Open sessions** section to Settings → Terminal that lists currently-open long-running zmx sessions and lets you kill them — aimed at catching sessions that "grow wild" (leaks from crashes, killed worktrees, or other instances).

- **Active in this window** — sessions tracked by this Alas window (from the registry). Kill routes through the normal `closeSession` path, so the registry stays consistent, and honors the existing "confirm before closing terminal tabs" setting.
- **Other / orphaned** — `alas-*` sessions in `zmx ls` not tracked here, badged **idle** (`clients == 0`) or **in use** (`clients ≥ 1`). Idle kills immediately; in-use kills require an explicit confirm, since they may belong to another Alas window (shared `ZMX_DIR`).
- **Kill all idle** — bulk-reaps every idle orphan behind a counted confirm.
- Loads on appear, after any kill, and via Refresh — no background polling. `zmx ls`/`kill` run off-main via the existing tracked-kill machinery.

### How it's built (layered)
- `ZmxSessionInfo` parser extended to capture the full `zmx ls` field set (`pid`, `clients`, `created`, `cmd`).
- `OpenSessionsClassifier` — a pure, unit-tested function splitting `zmx ls` output into active vs orphaned (idle/in-use), with `ZMX_SESSION_PREFIX` stripping and an explicit guard so a tracked tab reporting `clients=0` can never leak into "Kill all idle".
- `TerminalService.loadOpenSessions()` / `killOrphanSession(name:)` / `killIdleOrphans(_:)` wrappers.
- `OpenSessionsSection` SwiftUI view, wired into `TerminalPane`.

## Test Plan
- [x] `xcodebuild ... test` — full suite green (3476 tests, 0 failures)
- [x] Unit tests: parser (full + malformed fields), classifier (active/orphaned, idle/in-use, legacy/non-alas, daemon-died, prefix), service load + kill routing, `relativeAge`
- [ ] Settings → Terminal shows the **Open sessions** group between "Sessions" and "Startup scripts"
- [ ] An open tab appears under "Active in this window" with its working dir; killing it respects the confirm-close setting
- [ ] Orphaned `alas-*` sessions appear with idle/in-use badge + age; Refresh reloads; in-use kill prompts a confirm
- [ ] "Kill all idle" appears when ≥1 idle orphan exists and reaps them after confirm

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -709,6 +709,7 @@
 		BAE4D422FC3950E61FEE1D15 /* AppIntents.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = CA43AA9C8462D579E733B46C /* AppIntents.framework */; };
 		BAE6B1D93387807ADD7ED0CC /* ACPImageThumbnail.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5E1462C85D0E49127BCED81 /* ACPImageThumbnail.swift */; };
 		BB565B8EE9BA20EE961B32FB /* ThemeSidebarContrastTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B70ED4990428C46B69784970 /* ThemeSidebarContrastTests.swift */; };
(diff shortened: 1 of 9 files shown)


mrmans0n:
@codex review

PR #501, opened 2026-06-09
Fix remote tool call cards

mrmans0n:
## Summary
- Bust the remote web shell cache so mobile clients pick up the tool-card rendering fix.
- Render useful tool-call metadata when a tool invocation has no output content.
- Make remote tool-card row text explicit on mobile browsers.

## Validation
- xcodegen
- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test


mrmans0n:
@codex review

PR #504, opened 2026-06-10
fix(cursor-acp): fast mode persistence

mrmans0n:
## Summary
- Keep the selected ACP config-option value after a successful `session/set_config_option` call even when the agent response echoes stale state for that same option.
- Preserve refreshed dependent config options from the agent response.
- Add a regression test covering Cursor-style `fast=false` echo after selecting `true`.

## Root Cause
Cursor ACP can return a refreshed `configOptions` list whose `fast` option still contains the old current value. Alas previously replaced the optimistic local state with that response wholesale, so the Fast chip flipped back off shortly after the user enabled it.

## Validation
- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/ACPConfigOptionTests -quiet test`

Note: the full `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` run was attempted but hung locally and was interrupted after extended runtime.

### DIFF
--- a/Alas/Sources/ACP/Protocol/ACPConfigOption.swift
+++ b/Alas/Sources/ACP/Protocol/ACPConfigOption.swift
@@ -46,6 +46,28 @@ struct ACPConfigOption: Codable, Equatable, Identifiable, Hashable {
         try c.encodeIfPresent(currentValue, forKey: .currentValue)
         try c.encode(options, forKey: .options)
     }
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #507, opened 2026-06-11
fix(acp): bound ACP terminal output retention

mrmans0n:
## Summary

- Cap retained finished ACP terminal entries per chat session so old tool subprocess output cannot grow without bound.
- Include retained ACP terminal buffer bytes in debug memory diagnostics.
- Add regression coverage for finished terminal retention and diagnostic byte accounting.

## Validation

- `xcodegen`
- `git diff --check`

## Notes

Also ran `scripts/prune-xcode-state.sh --delete` locally to remove stale Alas/Ghostty DerivedData entries.


mrmans0n:
@codex review

PR #510, opened 2026-06-11
fix(worktree): speed up worktree selection

mrmans0n:
## Summary
- Defer and coalesce spaces persistence after worktree/space selection so the visual switch can render immediately
- Flush pending selection persistence on app termination
- Add regression coverage for deferred/coalesced selection persistence and flushing

## Verification
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/AppStateSpacesTests`
- `xcodegen`

Note: A full `xcodebuild ... test` run was attempted locally but stayed silent for several minutes and was interrupted.


mrmans0n:
@codex review

PR #526, opened 2026-06-15
fix(editor): find attachment

mrmans0n:
## Summary
- Bind the center-pane editor find controller through direct `CodeEditorView` lifecycle callbacks instead of same-view attach notifications.
- Refresh find matches as soon as the mounted `CodeTextView` is attached so searches run against the visible editor buffer.
- Add regression coverage for coordinator attach/detach callbacks.

## Root cause
`CodeEditorCoordinator.attach` posted `.codeEditorDidAttach` during `NSViewRepresentable.makeNSView`. The parent `EditorTabView` listened with `.onReceive`, which can miss notifications emitted before the SwiftUI subscription is installed. When that happened, the find controller never received the live `CodeTextView`, so searches reported no matches even though the text was visible.

## Verification
- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/EditorFindControllerTests -only-testing:AlasTests/EditorBufferTests test` (65 tests passed)

Note: full `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` was attempted, but the app-hosted test run hung after several minutes and was interrupted.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -5498,7 +5498,7 @@
 			repositoryURL = "https://github.com/apple/swift-markdown";
 			requirement = {
 				kind = upToNextMajorVersion;
-				minimumVersion = 0.4.0;
+				minimumVersion = 0.8.0;
 			};
 		};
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #527, opened 2026-06-15
Smooth diff review scrolling

mrmans0n:
## Summary
- Cache AppKit diff row geometry so repeated gutter/background drawing does not force fresh TextKit layout during scrolling
- Retain recently rendered review file sections through a wider scroll retirement band to avoid AppKit diff teardown/rebuild churn when reversing direction
- Add focused regressions for row-geometry reuse and render-window retention

## Verification
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/DiffReviewSurfaceTests -only-testing:AlasTests/DiffPaneViewTests -quiet test`
- `xcodegen`


mrmans0n:
@codex review

PR #535, opened 2026-06-16
fix(changes): stale changes snapshot flash

mrmans0n:
## Summary
- Invalidate cached right-pane snapshots when a worktree is created for an existing path/id.
- Render the right-pane loading skeleton until git refresh publishes trusted Changes/Files data.
- Add regression coverage for hiding stale cached changes after invalidation.

## Root cause
The right pane could render directly from a cached `RightPaneState.changes` snapshot while a new worktree refresh was still pending. If a path/id was reused, the UI could briefly show old diff counts and tree rows before `git status` completed.

## Validation
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/RightPaneStateBaseBranchTests`
- `xcodegen`

Note: the full `xcodebuild ... test` command was started locally but hung in the broader terminal/ACP test area after many tests had passed, so it was interrupted.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -5498,7 +5498,7 @@
 			repositoryURL = "https://github.com/apple/swift-markdown";
 			requirement = {
 				kind = upToNextMajorVersion;
-				minimumVersion = 0.4.0;
+				minimumVersion = 0.8.0;
 			};
 		};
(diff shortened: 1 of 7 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #536, opened 2026-06-16
fix(review): improve review pane interactions

mrmans0n:
## Summary

- keep the existing review diff visible during background reloads so the pane no longer flashes to a full-screen loading state
- restyle the inline draft/edit comment composer and add Cmd+Enter to save plus Escape to cancel
- show a compact plus affordance when hovering commentable diff gutter rows

## Validation

- `xcodegen`

Full local suite result: 3949 tests in 438 suites passed.


mrmans0n:
@codex review

PR #537, opened 2026-06-16
feat(sidebar): house glyph for the main worktree row

mrmans0n:
## Summary

- The main worktree's sidebar row now shows an outline **house** glyph (SF Symbol `house`, tinted `fg-muted`) instead of the branch glyph (`fg-faint`). Feature worktrees are unchanged.
- The cue marks the worktree's **role** — the primary on-disk checkout — which is independent of whatever branch it has checked out (so it stays correct even when main isn't on `main`, and avoids a redundant "main" label).
- Wiring: `Icon.symbol("home") → "house"`; a public `ProjectsManager.isMain(_:in:)` wraps the existing path-comparison used to pin main at position 0; `SidebarView` → `RepoGroupView` → `WorktreeRowView` thread an `isMain` flag that selects the glyph + tint. No data-model, theme-token, or layout changes.

## Test Plan

- [x] `IconSymbolTests` — `home` resolves to `house`, `branch` unchanged
- [x] `ProjectsManagerTests` — `isMain` true for the primary checkout, false for a real linked feature worktree
- [x] `SidebarVisibilityTests` + layout suites green; app builds clean
- [ ] Visual: main row (position 0) shows the house at `fg-muted`; feature rows still show the branch glyph at `fg-faint`; selection bar / metrics / harness pill unchanged


mrmans0n:
@codex review

PR #551, opened 2026-06-17
feat(review): vendor brand glyphs for code host headers, drop focus ring

mrmans0n:
## What

The review-readiness drawer header showed a generic branch glyph and drew the default macOS focus ring around its collapsed row. This swaps in real vendor brand marks and removes the focus outline so the header matches the rest of the chrome.

## Changes

- Ship template SVGs for **GitHub** and **GitLab** in `CodeHostLogos/` (official monochrome marks) so they tint with the resolved `Icon` color.
- Route `Icon` to load those assets when the name matches `github`/`gitlab`, and drop the `circle.hexagongrid` fallback that previously stood in for GitHub.
- Expose `CodeHostKind.iconName` and thread it through `ReviewReadinessModel` so the header picks the right brand mark, falling back to the branch glyph when no host is known.
- Disable the focus ring on the collapsed row of the review loop drawer.

## Notes

GitHub remotes show the Octocat, GitLab remotes show the tanuki. Both are template-rendered, so they pick up the same faint tint as the surrounding glyphs.


mrmans0n:
@codex review

PR #552, opened 2026-06-17
fix(editor): stale markdown preview tab content

martinbirn:
## Summary
- Key markdown preview render results by worktree/tab/path identity so preview-only tabs cannot reuse stale content from a previous SwiftUI view slot.
- Invalidate and re-render immediately when the active markdown render identity changes.
- Reject late async render completions after another markdown tab has become active.
- Add unit coverage for the markdown preview cache identity invariant, including stale slot reuse and late completion behavior.

## Verification
- Manual: reproduced the original issue with multiple preview-only markdown tabs in the traveler repo, then confirmed switching tabs updates preview content immediately after the fix.
- Passed: `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/MarkdownViewModeTests test`
- Passed: `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build`
- Passed: `git diff --check`
- Attempted: `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test`; interrupted after ~25 minutes with the run stuck at existing `ContentSearcherTests.findsMatchesAcrossFiles()`. Sample captured locally at `/tmp/alas-full-test-hang-sample.txt`.

Note: `xcodegen` was not run because `project.yml` was unchanged and tests were added to an existing project file already present in `Alas.xcodeproj`.

### DIFF
--- a/Alas/Sources/Code/Markdown/MarkdownTabView.swift
+++ b/Alas/Sources/Code/Markdown/MarkdownTabView.swift
@@ -14,8 +14,17 @@ struct MarkdownTabView: View {
     let onRevealInFiles: (String) -> Void
     @Environment(\.theme) var theme
 
-    @State private var renderResult: MarkdownRenderResult?
+    @State private var renderCache = MarkdownPreviewCache<MarkdownRenderResult>()
     @State private var debounceTask: Task<Void, Never>?
+
+    private var renderIdentity: MarkdownRenderIdentity {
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #553, opened 2026-06-18
feat(diff): keep diff whitespace hidden by default

mrmans0n:
## Summary
- keep the diff whitespace/paragraph toggle as local per-screen state
- leave layout and wrap preferences persisted
- add regression coverage for local whitespace defaults

## Verification
- `rtk xcodegen`
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/DiffPaneViewTests test`

Full `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` was run twice and both runs failed only in `GitHubCLIProviderVerdictTests.startReviewFetchesPRNodeIDThenCallsAddPullRequestReview` with `headSHA unavailable`; that test passed in isolation.


mrmans0n:
@codex review

PR #554, opened 2026-06-18
fix(remote): web tool rows on mobile

mrmans0n:
## Summary
- Render remote web tool rows with an ARIA button div instead of a native button to avoid mobile Safari dropping the collapsed row contents.
- Bump remote web JS/CSS cache keys so clients load the updated renderer.
- Add a remote web asset regression test for the mobile tool-row renderer.

## Verification
- `node --check Alas/Resources/RemoteWeb/app.js`
- `xcodegen`
- `git diff --check`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/RemoteWebAssetTests`
- Full `xcodebuild ... test` was run and failed with the existing order-dependent `GitHubCLIProviderVerdictTests/startReviewFetchesPRNodeIDThenCallsAddPullRequestReview()` `headSHA unavailable`; rerunning that single test passed.

### DIFF
--- a/Alas/Resources/RemoteWeb/app.js
+++ b/Alas/Resources/RemoteWeb/app.js
@@ -1532,6 +1532,13 @@ function singleLine(value) {
   return (value || "").replace(/\s+/g, " ").trim();
 }
 
+function handleCardToggleKeydown(event) {
+  if (event.key !== "Enter" && event.key !== " ") return;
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #577, opened 2026-06-19
feat(acp): context ring tooltip and click target

mrmans0n:
Shows the exact context-window percentage on hover and makes the ring itself clickable to open the usage details popover.


mrmans0n:
@codex review

PR #581, opened 2026-06-19
fix(acp): discard replay chunks that arrive while transcript is idle

mrmans0n:
## Summary

- Late-arriving ACP replay updates (a race between the server committing updates and the client capturing `yieldedUpdateCount` in the suppression window) were hitting `completedOutputBoundaryMessageIds` on the previous turn's message and creating a duplicate `N+1` bubble with identical content — visible as two back-to-back agent messages with the same text.
- Fix: in `appendStreaming`, guard on `streamingState != .idle` before allowing a boundary-marked message to trigger a new message. The runner always transitions to `.sending` before a legitimate new-turn chunk arrives, so this only drops replay slips.
- Update 4 existing boundary tests to set `.sending` before new-turn chunks (accurately reflecting the production invariant) and add a regression test for the replay-slip scenario.

## Test plan

- [ ] Existing `ACPSessionTests` boundary tests pass with the `.sending` state set correctly
- [ ] New `replayChunkWhileIdleIsDiscarded` test asserts the duplicate is prevented
- [ ] Full CI suite green

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSession.swift
+++ b/Alas/Sources/ACP/Session/ACPSession.swift
@@ -85,6 +85,15 @@ final class ACPSession: ObservableObject, Identifiable {
     private static let metadataPreviewLimit = 4096
     private var contextRecoveryExpiryTask: Task<Void, Never>?
 
+    /// When false, `appendStreaming` discards chunks that would cross a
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #604, opened 2026-06-22
fix(editor): preserve horizontal offset when revealing files

mrmans0n:
## Summary
- Preserve the editor scroll view's existing horizontal offset when applying reveal targets.
- Add a regression test for opening/revealing a far-right match without losing the leading edge.

## Testing
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination platform=macOS,arch=arm64 -only-testing AlasTests/EditorBufferTests/coordinatorRevealKeepsEditorHorizontallyAtLeadingEdge CURRENT_ARCH=arm64 ARCHS=arm64 ONLY_ACTIVE_ARCH=YES test -quiet`


mrmans0n:
@codex review

PR #606, opened 2026-06-22
fix(editor): keep line-number gutter clear at leading edge

mrmans0n:
Respect AppKit's negative clip origin when a vertical ruler reserves
space for line numbers, so fully-left editor scrolls no longer hide
the first characters under the gutter.


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #620, opened 2026-07-03
fix(acp): clear stranded transcript-restore spinner when steered

mrmans0n:
## Problem

After restoring a session from transcript, the "Restoring model context from transcript…" spinner sometimes never goes away.

## Root cause

The `.sendingTranscript` recovery status is cleared **only** by the completion callback wired in `ACPSessionManager.sendTranscriptAsContext`. But `ACPSessionRunner.sendRecoveryContext` skipped that callback entirely whenever a newer prompt had taken over the transport (`hasNewerActivePrompt`). When the user **steers** a new prompt while the recovery context is still in flight (the composer even shows "⌥↵ to steer"), the steer installs a newer `activePromptID`, the recovery callback never fires, and the spinner strands forever. It's a race between the recovery RPC completing and the steer's replacement prompt installing itself — hence "sometimes."

The `!hasNewerActivePrompt` guard is correct for `sendNow` (a superseded turn legitimately hands its UI state to the successor), but wrong for the recovery status, which has no successor to hand off to and must always resolve.

## Fix

In `sendRecoveryContext`, always invoke `onCompleted` (both success and error paths). The callback only touches recovery status / warning / persistence — never streaming or queue state — so it can't stomp the newer prompt. Under a steer the recovery prompt is marked cancelled, so the status now resolves to `.failed` (actionable, with Retry, and keeps the recovery-pending flag so it retries later) instead of a phantom spinner.

## Test

`recoveryContextSupersededBySteerClearsStatus` deterministically gates the recovery prompt in flight, steers a replacement that takes over the transport, then releases the recovery RPC and asserts the status leaves `.sendingTranscript`. It times out before the fix and passes after. The full attach-restore suite (30 tests) and the runner/lease suites (59 tests) pass.

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSessionRunner.swift
+++ b/Alas/Sources/ACP/Session/ACPSessionRunner.swift
@@ -1180,30 +1180,32 @@ extension ACPSessionRunner {
                 await MainActor.run {
                     let wasCancelled = self.cancelledPromptIDs.remove(promptID) != nil
                     let isActivePrompt = self.activePromptID == promptID
-                    let hasNewerActivePrompt = self.activePromptID != nil && !isActivePrompt
                     if isActivePrompt {
                         self.activePromptID = nil
                         if self.deferCompletedOutputBoundaryUntilUpdatesDrain() {
                             self.flushQueueIfIdle()
                         }
                     }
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #624, opened 2026-07-03
fix(acp): audit and harden JSON-RPC cancellation teardown

### LINKED ISSUE #614: Audit ACP JSON-RPC cancellation handling
Recent Claude and Codex ACP adapters added explicit ACP request cancellation handling.

Alas already supports `session/cancel` for active prompts, but should be audited for JSON-RPC request cancellation semantics around pending `session/prompt`, permission, question, file, terminal, and config-option requests.

Acceptance criteria:
- Confirm whether Alas should send ACP cancellation notifications for abandoned in-flight requests.
- Ensure local prompt cancellation, tab detach, runner shutdown, and takeover do not leave adapter-side work hanging.
- Add focused tests for any missing cancellation paths.

References:
- `agentclientprotocol/claude-agent-acp` June 23, 2026: handle ACP request cancellation signals.
- `agentclientprotocol/codex-acp` June 25, 2026: handle ACP request cancellation.

mrmans0n:
## Summary

Audit and fix of ACP JSON-RPC request-cancellation handling. Alas is the ACP client; `session/cancel` was already sent correctly. The audit found the real gaps were teardown continuation hygiene — pending continuations that could be left suspended, hanging callers and leaking the runner object graph.

Fixes #614.

## Findings addressed

- **Deterministic pending-drain on shutdown.** `ACPStdioClient.shutdown()` only drained its outbound `pending` request map via the transport's `.exited` event, which dispatch-task cancellation could race away — leaving an in-flight `session/prompt` continuation unresumed and leaking the runner → session → connection graph. `shutdown()` now drains `pending` itself via a drain-once `drainPending(with:)` helper (guarded by `stateLock` + a `didDrainPending` flag), with the `.exited` handler kept idempotent. `send()` gains a fast-fail guard after drain and only resumes on transport-send failure if it still owned the pending entry (removes a double-resume → process-trap window).
- **Permission-continuation cleanup on runner stop.** `ACPSessionRunner.stop()` now calls `policy.userCancelled()`, so a detach/takeover landing while a permission prompt is parked can't strand `ACPPermissionPolicy.pendingContinuation` (previously only `userCancel()` resolved it).

## Audit conclusions (acceptance criteria)

- **No extra ACP cancellation notifications needed for abandoned in-flight requests.** User cancel already sends `session/cancel`; every other teardown path (`detach`, `stop`, takeover/`standDown`) calls `connection.shutdown()`, which SIGTERMs the adapter process, so adapter-side work is reaped by process death. The defect was our own continuations not being freed, not missing messages.
- **No RPC timeout added** — a timer is a fragile proxy for a teardown-correctness bug and would risk false-killing legitimately long `session/prompt` turns; deterministic drain-on-teardown fixes it properly.
- SIGKILL escalation in `terminate()` left out of scope (process-lifecycle, distinct from JSON-RPC cancellation).

## Tests

- `FakeJSONRPCTransport` gains an `emitExitOnTerminate` knob so a test can mimic the real async-SIGTERM transport (the synchronous `.exited` previously masked the bug).
- New regression tests: shutdown drains an in-flight request without `.exited`; drain is idempotent with a transport that also emits `.exited`; `ACPConnection.shutdown()` unwinds an in-flight prompt; `stop()` resolves a parked permission continuation as cancelled.
- A `boundedResult` test helper abandons (rather than awaits) a hung task so a future regression fails fast instead of hanging the run.

Focused suites pass locally (`ACPStdioClientTests` 4/4, `ACPSessionRunnerTests` 32/32); full suite runs in CI.

### DIFF
--- a/Alas/Sources/ACP/Protocol/ACPStdioClient.swift
+++ b/Alas/Sources/ACP/Protocol/ACPStdioClient.swift
@@ -224,9 +224,13 @@ final class ACPStdioClient: ACPClient, @unchecked Sendable {
                 try transport.send(body)
             } catch {
                 stateLock.lock()
-                pending.removeValue(forKey: id)
+                // If a concurrent drainPending (shutdown()/`.exited`) already
+                // removed and resumed this entry, `removeValue` returns nil
+                // here — in that case `cont` has already been resumed and we
+                // must not resume it again (that would trap).
+                let stillPending = pending.removeValue(forKey: id) != nil
                 stateLock.unlock()
-                cont.resume(throwing: error)
+                if stillPending { cont.resume(throwing: error) }
             }
         }
         return ACPResponse(body: data)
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #628, opened 2026-07-03
fix(remote): stop tool cards collapsing to 1px as transcript fills

mrmans0n:
## Problem

In the remote web app chat screen, tool cards render correctly at first but as the transcript fills the screen they shrink progressively until they are ~1px-tall lines — visible but useless. Plain text messages are unaffected.

## Root cause

`#messages` is a flex column:

```css
#messages { ... display: flex; flex-direction: column; }
```

Its children carry the default `flex-shrink: 1`, so once the total content height exceeds the container the flex algorithm shrinks them to fit. Tool cards (`.m-tool`) use `overflow: hidden`, which per the flexbox spec gives a flex item an **automatic minimum size of 0** — so they collapse all the way down. Text messages keep their content-based minimum size, which is why only tool cards degrade.

## Fix

Pin every message row so it keeps its natural height and lets `#messages` scroll instead:

```css
#messages > * { flex-shrink: 0; }
```

Bumped the `style.css` cache-buster (`v=33` → `v=34`) and the service-worker cache name (`v24` → `v25`) so clients pick up the new stylesheet, and updated the asset-version tests.

## Tests

Added `messageRowsDoNotShrinkInTranscriptFlexColumn` to `RemoteWebAssetTests` and updated the existing cache-buster assertions.

> Note: touches the same versioned assets as #626; whichever merges second will need a trivial cache-buster rebase.

### DIFF
--- a/Alas/Resources/RemoteWeb/index.html
+++ b/Alas/Resources/RemoteWeb/index.html
@@ -11,7 +11,7 @@
   <link rel="icon" href="/icon.svg" type="image/svg+xml" />
   <link rel="apple-touch-icon" href="/icon-180.png" />
   <title>Alas Remote</title>
-  <link rel="stylesheet" href="/style.css?v=34" />
+  <link rel="stylesheet" href="/style.css?v=35" />
 </head>
 <body>
   <header id="bar">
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

PR #631, opened 2026-07-04
feat(changes): default main worktree comparison to origin/baseBranch

mrmans0n:
# Main Worktree Defaults to Remote Base Branch in Changes Tab

## Summary

When a worktree's current branch matches the configured base branch, the Changes tab's Commits section currently compares HEAD against itself (`main..main`), producing an empty and useless list. This change makes the default comparison target `origin/<baseBranch>` in that case, so the main worktree shows meaningful ahead/behind information.

## Motivation

The configured base branch (`AppConfig.worktrees.baseBranch`) is used as the default comparison ref for every worktree. For a feature worktree checked out on `feature/x` with `baseBranch = "main"`, comparing against `main` is correct. But for the primary worktree checked out on `main` itself, the comparison becomes `main..main`, which is empty. Users expect the main worktree to show how far ahead/behind `origin/main` they are.

### Behavior

- In `RightPaneStore.state(for:baseBranch:trackUpstreamForCommits:)`, before seeding a new `RightPaneState`, compute an **effective default base branch** for the worktree:
  - If `worktree.branch` equals `baseBranch`, prefer `origin/<baseBranch>`.
  - If `origin/<baseBranch>` does not exist in the worktree, fall back to `baseBranch`.
  - If `worktree.branch` differs from `baseBranch`, keep `baseBranch`.
- Detached HEAD or empty branch names keep the configured `baseBranch` unchanged.
- If `baseBranch` already contains a remote qualifier (e.g. `upstream/main`), the equality check (`worktree.branch == baseBranch`) will not match ordinary local branch names, so no additional `origin/` prefix is prepended.

### Data Flow

```
AppState / RightPaneView asks RightPaneStore for state
  → RightPaneStore computes effectiveBaseBranch(worktree, baseBranch)
    → synchronous: returns origin/<baseBranch> or baseBranch
    → asynchronous: optionally verifies origin/<baseBranch> resolves
  → new RightPaneState seeded with effective base branch
  → lastConfigBaseBranch tracks effective default (not raw config string)
  → Settings change resets effective default
  → User override via BaseBranchSelector still preserved
```

### Component Architecture

- **`RightPaneStore`** — Gains a small helper `effectiveBaseBranch(worktree:baseBranch:)` that implements the rule above. The existing `state(for:baseBranch:trackUpstreamForCommits:)` uses its output as the initial `baseBranch` and `lastConfigBaseBranch`.
- **`RightPaneState`** — No changes to public interface. Its `baseBranch` simply starts at the resolved effective default.
- **`BaseBranchSelector` / `CommitsSectionView`** — No changes. They already operate on `rps.baseBranch` and `rps.comparisonRef`.

### Concurrency

The remote ref existence check is done asynchronously to avoid blocking UI construction. The initial state is created with the synchronous fallback (`baseBranch`). If an async probe confirms `origin/<baseBranch>` resolves, and the user has not yet overridden the base branch, the store updates the cached state's `baseBranch` and `lastConfigBaseBranch` and triggers a refresh.

### Error Handling

- If `origin/<baseBranch>` cannot be resolved (no remote named origin, missing remote ref), the fallback `baseBranch` remains. The UI behaves exactly as today.
- Probe failures are logged via `os.Logger` and do not surface user-facing errors.

## Testing Plan

Add Swift Testing unit tests for `RightPaneStore.effectiveBaseBranch` covering:

- `main` worktree + `baseBranch = "main"` resolves to `"origin/main"` when the remote ref exists.
- Falls back to `"main"` when `origin/main` does not exist.
- `feature` worktree + `baseBranch = "main"` keeps `"main"`.
- Manual user override is not overwritten by an async probe.
- Detached HEAD / empty branch keeps `"main"`.

## Out of Scope

- Persisting the effective default per worktree. It is recomputed each time the state is created or the config changes.
- Changing the Base Branch selector UI or the global Settings default.
- Affecting the Review Changes center pane or review-request flows; they have their own scope/branch resolution.

## Risks / Open Questions

- **Remote named something other than `origin`:** The rule hard-codes `origin/<baseBranch>`. Repos that use a different primary remote name will fall back to the configured base branch. This matches the existing convention in `BaseBranchSelector.smartList` and `GitService.resolveBaseRef`.
- **No network fetch:** We only check whether the remote ref exists locally (`refs/remotes/origin/<baseBranch>`). We do not fetch. This is intentional and consistent with the existing sync status behavior.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -294,6 +294,7 @@
 		4135B5BC24856ED442E39069 /* GitServiceStashTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1B218326D14AA360C7476D43 /* GitServiceStashTests.swift */; };
 		4151B68B1CF0A947AEC4C00A /* ACPAuthNudgeBannerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F68B06E93430D1C7B02F6BB2 /* ACPAuthNudgeBannerTests.swift */; };
 		4153DAD75F27395D2D385BED /* LanguageServerAvailabilityTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A8CC42BD54479E057902D392 /* LanguageServerAvailabilityTests.swift */; };
+		4186372EBF16A9C7CB38F558 /* RightPaneStoreBaseBranchTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8451C5BCEE50F118F07CDD83 /* RightPaneStoreBaseBranchTests.swift */; };
 		418D526733AA6B50A66CCCAF /* TabsManagerReviewRequestDraftTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 83A8622BA130E8FA339EDCEB /* TabsManagerReviewRequestDraftTests.swift */; };
 		419029F5618BC364426491E5 /* ProjectIcon.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDEB760B98A1482CF30140A4 /* ProjectIcon.swift */; };
 		41C3F07661CA9B2B4E2C7BB1 /* GhosttyConfigBuilder.swift in Sources */ = {isa = PBXBuildFile; fileRef = BCA505FF2E6CFA8278C95C6E /* GhosttyConfigBuilder.swift */; };
@@ -1720,6 +1721,7 @@
 		82F696636690652C48CE87D9 /* AlasGhostty.Config.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AlasGhostty.Config.swift; sourceTree = "<group>"; };
 		835D847AD52A2CEB36DD5022 /* LegacyHookSweepTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LegacyHookSweepTests.swift; sourceTree = "<group>"; };
 		83A8622BA130E8FA339EDCEB /* TabsManagerReviewRequestDraftTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TabsManagerReviewRequestDraftTests.swift; sourceTree = "<group>"; };
+		8451C5BCEE50F118F07CDD83 /* RightPaneStoreBaseBranchTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RightPaneStoreBaseBranchTests.swift; sourceTree = "<group>"; };
 		84A01D6512394EF1C41782AD /* ACPMarkdownInlineTextView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ACPMarkdownInlineTextView.swift; sourceTree = "<group>"; };
 		84E56989C04A4A478D68084C /* ACPMarkdownBlockCacheTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ACPMarkdownBlockCacheTests.swift; sourceTree = "<group>"; };
 		8526C9DB63760BDD2CE998BD /* fs-write.json */ = {isa = PBXFileReference; lastKnownFileType = text.json; path = "fs-write.json"; sourceTree = "<group>"; };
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex I added markSnapshotUnknown() before applying the fallback in the async probe, which bumps snapshotInvalidationGeneration so any refresh still in flight from the initial origin/base guess is discarded when it completes. Could you re-review?

mrmans0n:
@codex address that feedback

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
Pushed cacfce0d which captures requestedConfig before mutating lastConfigBaseBranch and uses it in the probe guard, plus defers state.start() for new states until verification finishes. This should resolve all three threads: the verified fallback is remembered across renders, probes don't rearm, and the initial refresh doesn't race the fallback.

mrmans0n:
@codex review

mrmans0n:
@codex I removed the standalone activate() method and inlined the activation logic, making the deferStart guard straightforward: when the effective default is a guessed origin ref, state.start() is skipped at activation and is instead called inside the probe task after the ref is confirmed or after the fallback is applied. When the default is not a guess, start() runs immediately as before. This avoids the race entirely for new states.

mrmans0n:
@codex review — pushed ebd1208b. Key changes since the last review:

1. `lastConfigBaseBranch` now always records the configured default (`origin/<base>` when on base branch), not the resolved fallback. So after falling back to `<base>`, the next render compares `lastConfigBaseBranch == rawDefault` and sees no settings change, preventing the rearm loop.
2. New-state activation no longer calls `start()` when the effective default is a guessed origin ref; `start()` is deferred to the probe completion path, and only runs if the state is still active. This removes the race with the initial refresh.

Updated the test to assert the new `lastConfigBaseBranch` semantics. All 10 `RightPaneStoreBaseBranchTests` pass locally.

mrmans0n:
@codex review — I resolved all outstanding review threads because the current commit (ebd1208b) addresses them:

- Fallback persistence: `lastConfigBaseBranch` now records the configured default, not the resolved fallback, so repeated renders no longer look like Settings changes.
- Stale refresh race: new-state activation defers `start()` until the origin-ref probe finishes; existing-state fallback calls `markSnapshotUnknown()` before applying the fallback.
- Deferred initial refresh: `result.start()` is skipped when `shouldDeferInitialRefresh` is true.

Please take a fresh look and let me know if anything remains.

mrmans0n:
@codex review — pushed 5aa86c26. This addresses both remaining threads:

1. **Track raw settings value for config changes:** `lastConfigBaseBranch` now stores the raw configured base branch (e.g. `main`), while `baseBranch` holds the effective comparison ref. A Settings change from `main` to `origin/main` is now detected correctly. I also added an `effectiveChanged` branch so that worktree branch changes (which alter the effective default without changing the raw config) still update `baseBranch`.
2. **Deactivate/reactivate race:** Added `isAwaitingBaseBranchProbe` to `RightPaneState`. New deferred states set it; reactivation checks it and skips `start()` while the probe is still running. The probe task clears the flag and only starts the state if it is still active.

Also added `baseBranchProbeTask` and cancellation so stale probe results never overwrite a newer default. All 10 targeted tests pass.

mrmans0n:
@codex review — pushed 503b518b. I added `lastEffectiveBaseBranch` to track the last effective default the store applied, separate from `lastConfigBaseBranch` (raw settings) and `baseBranch` (current comparison ref, possibly a verified fallback). The `effectiveChanged` branch now compares against `lastEffectiveBaseBranch` instead of `baseBranch`, so a verified fallback (`main`) is no longer mistaken for a worktree-branch change on the next render. Repeated renders with the same effective default keep the fallback. Targeted tests still pass.

mrmans0n:
@codex review — pushed 9d67d423. I separated the comparison base from the review-loop base:

- `RightPaneState` is now initialized with the configured base branch (e.g. `main`) for `ReviewLoopState`, then `state.baseBranch` is overwritten to the effective comparison default (`origin/main`).
- On Settings change, `reviewLoop.updateBaseBranch(baseBranch)` uses the raw configured branch, not `origin/<base>`.
- The async probe that verifies/falls back the comparison ref no longer touches `reviewLoop` at all; it only mutates `state.baseBranch`.
- Effective worktree-branch changes also leave the review-loop base untouched.

Targeted tests still pass.

mrmans0n:
@codex review — pushed 988c7cac. I exposed `ReviewLoopState.currentBaseBranch` and updated `RightPaneState.refreshReviewLoop` to build `ReviewLoopLocalState` with that value instead of `state.baseBranch`. This decouples the review-loop base (configured/user-selected) from the commits comparison ref (effective default/verified fallback) throughout the refresh path. Targeted tests pass.

PR #632, opened 2026-07-04
fix(acp): transcript layout after pane resize

mrmans0n:
## Summary

- Restore ACP transcript scroll anchoring when the message-list viewport width changes.
- Keep tail-following sessions pinned after resize-induced wrapping/content-height changes.
- Add focused coverage for width-change tail restore policy and hosted inline markdown reflow.

## Validation

- `rtk xcodegen`
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' ALAS_FFF_TARGET_ARCH=arm64 -quiet build`
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' ALAS_FFF_TARGET_ARCH=arm64 -only-testing:AlasTests/ACPMessageListPaginationTests -only-testing:AlasTests/ACPMarkdownInlineRendererTests test`
- Full suite was also run with the same arm64 fff override; it failed in two order-dependent `ACPSessionTests` for long image/data URI tool-call assets, and those two tests passed when rerun isolated.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPMessageList.swift
+++ b/Alas/Sources/ACP/UI/ACPMessageList.swift
@@ -223,6 +223,16 @@ struct ACPMessageList: View {
                     restoreTailIfNeeded(proxy: proxy, animated: false)
                     restoreRememberedAnchorIfNeeded(proxy: proxy)
                 }
+                .onChange(of: viewport.size.width) { oldWidth, newWidth in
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #637, opened 2026-07-05
fix(remote): Clarify remote QR address selection

mrmans0n:
## Summary

- add a dedicated Remote settings row for choosing which advertised address is encoded into the pairing QR
- make advertised address rows copy-only so QR selection lives in one place
- disambiguate duplicate segmented-picker labels and give the hidden picker a useful accessibility label

## Verification

- `env ALAS_FFF_TARGET_ARCH=arm64 ALAS_ZMX_TARGET_ARCH=arm64 rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' -quiet build`


mrmans0n:
@codex review

PR #638, opened 2026-07-05
fix(acp): measure chat width during task resize

mrmans0n:
## Summary
- measure ACP chat content width from the actual chat column geometry during task sidebar transitions
- keep the plan sidebar width in the shared chat layout constants
- add focused coverage for measured chat-column width and repeated inline markdown resize bounds

## Verification
- `rtk xcodegen`
- `ALAS_FFF_TARGET_ARCH=arm64 rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet test -only-testing:AlasTests/ACPChatLayoutTests -only-testing:AlasTests/ACPMarkdownInlineRendererTests`

## Notes
- The documented build without `ALAS_FFF_TARGET_ARCH=arm64` still fails locally because the fff build phase cannot find the `x86_64-apple-darwin` Rust target.
- A full local test run with `ALAS_FFF_TARGET_ARCH=arm64` completed with unrelated failures in `DiffPaneViewTests/lineNumberRulerKeepsReviewSelectionForNonExpandableRows` and two `ACPSessionTests` raw-output asset tests.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPChatLayout.swift
+++ b/Alas/Sources/ACP/UI/ACPChatLayout.swift
@@ -3,6 +3,7 @@ import CoreGraphics
 enum ACPChatLayout {
     static let defaultContentMaxWidth: CGFloat = 720
     static let wideContentMaxWidth: CGFloat = 960
+    static let planSidebarWidth: CGFloat = 320
 
     static func chatColumnWidth(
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #640, opened 2026-07-05
Improve diff comment contrast

mrmans0n:
## Summary

- Render syntax comments in added/deleted diff rows with the main code foreground instead of the faint comment color.
- Apply the same rule to the selectable diff text builder so diff components stay visually consistent.
- Add regression coverage for changed-row comment foreground color.

## Why

The faint comment syntax color became hard to read over the more vibrant green added-line background in PR review diffs.

## Validation

- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' CURRENT_ARCH=arm64 ALAS_FFF_TARGET_ARCH=arm64 -only-testing:AlasTests/DiffSelectableTextTests test`
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' CURRENT_ARCH=arm64 ALAS_FFF_TARGET_ARCH=arm64 -quiet build`

### DIFF
--- a/Alas/Sources/Center/Diff/DiffCodeText.swift
+++ b/Alas/Sources/Center/Diff/DiffCodeText.swift
@@ -61,6 +61,7 @@ struct DiffCodeText: View {
             to: output,
             text: text,
             fileExtension: fileExtension,
+            inlineTone: inlineTone,
             theme: theme,
             visibleLength: visibleLength
         )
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #643, opened 2026-07-07
fix(center): cache resolved code font to prevent beachballs

mrmans0n:
## Summary
- A user reported a beachball; the hang report showed the main thread stuck in `CenterTypography.resolveCodeFont` → `NSFontManager.availableMembers(ofFontFamily:)`.
- `resolveCodeFont` is called once per visible code line (diff, merge, markdown surfaces). With a custom code font configured, every call re-enumerated all system font families via `NSFontManager`, an expensive AppKit call. On a large diff this compounded into a multi-second main-thread stall.
- Cache the resolved `NSFont` by `(family, size)` so the expensive lookup happens once per distinct font configuration instead of once per rendered line.

## Test plan
- [x] `xcodegen` + `xcodebuild ... build` succeeds
- [x] `xcodebuild ... test -only-testing:AlasTests/DiffPaneViewTests` — same result before/after the change (one pre-existing, unrelated failure)

### DIFF
--- a/Alas/Sources/Center/CenterTypography.swift
+++ b/Alas/Sources/Center/CenterTypography.swift
@@ -23,10 +23,28 @@ enum CenterTypography {
         (lineHeightMultiple - 1) * size * 1.2
     }
 
+    /// `resolveCodeFont` is called for every visible line of code across the diff,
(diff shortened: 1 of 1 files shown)


mrmans0n:
@codex review

PR #645, opened 2026-07-07
fix(updates): run brew update before upgrading the cask

mrmans0n:
## Summary
- The update dialog's "Update" action ran `brew upgrade --cask mrmans0n/tap/alas` directly, but the app's own update check queries GitHub releases (not brew), so a stale local Homebrew tap index could make the upgrade silently no-op even after the app correctly detected a new release.
- `SelfUpdateCommand.homebrew` now chains `brew update && brew upgrade --cask mrmans0n/tap/alas` via `sh -c`, refreshing the tap before upgrading. The copy-pasteable command text shown in the update sheet is updated to match.

## Test plan
- [x] `AlasTests/Updates/SelfUpdaterTests.swift` updated and passing locally
- [ ] CI green

### DIFF
--- a/Alas/Sources/Updates/SelfUpdater.swift
+++ b/Alas/Sources/Updates/SelfUpdater.swift
@@ -8,22 +8,40 @@ struct SelfUpdateCommand: Equatable, Sendable {
 
     let executable: String
     let arguments: [String]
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #649, opened 2026-07-08
fix(files): stop compacted directory rows blinking on refresh

mrmans0n:
## Problem

The Files tab flattens single-child directory chains into one row (e.g. `src/main/test` instead of `src` → `main` → `test`). Lazily-loaded directories — ignored/excluded roots and anything expanded on demand — blinked on every refresh: the row briefly showed the uncompacted tree, then re-collapsed into the flattened row.

## Root cause

Two compounding issues:

1. **Refresh discarded expanded lazy subtrees.** `refresh()` replaced `fileTree` with a freshly built tree where lazy directories come back as `.notLoaded` with no children. Any directory the user had expanded collapsed, and the auto-loader then re-fetched it level by level, visibly re-expanding and re-collapsing — on every FS-watcher refresh, unprompted.

2. **Unstable row identity.** The compacted row's `.id` was keyed on the chain's *terminal* path, which moves deeper as levels load. SwiftUI treats an `.id` change as a new identity, so it tore the row down and rebuilt it on each growth step (a hard flash) instead of updating the label in place.

## Fix

- `preservingLazyChildren` grafts previously loaded lazy subtrees into the freshly built tree on refresh, so expanded/compacted chains stay stable.
- `loadFileTreeChildren` skips the intermediate `.loading` flip when reconciling an already-loaded directory, keeping it in its compacted chain while contents refresh in the background.
- The compacted row is keyed on the stable chain **root** instead of the terminal, with zero-size anchors for the remaining chain paths so scroll-to-reveal still resolves any of them.

## Tests

Added unit tests for `preservingLazyChildren` (graft onto not-loaded dir, graft nested lazy subtree, leave unmatched/unloaded dirs untouched) and `fileTreeNode(at:in:)`. Existing compact-chain, merge, and reveal tests still pass.

### DIFF
--- a/Alas/Sources/Right/FilesTabView.swift
+++ b/Alas/Sources/Right/FilesTabView.swift
@@ -135,9 +135,15 @@ struct FilesTabView: View {
                         }
                         .contentShape(Rectangle())
                         .background(chain.chainPaths.contains(revealPath ?? "") ? theme.color("bg-hover") : Color.clear)
-                        .id("dir:\(terminal.path)")
+                        // Key the row on the chain's root (stable) rather than its
+                        // terminal, which moves deeper as levels load. A stable id
+                        // lets SwiftUI update the label in place while the chain
+                        // compacts instead of tearing the row down and rebuilding
+                        // it (a visible flash). Remaining chain paths get zero-size
+                        // anchors so scroll-to-reveal still works for any of them.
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #652, opened 2026-07-09
fix(diff): Auto-focus the diff pane comment composer

mrmans0n:
On the staged/unstaged diff pane, leaving a comment now focuses the comment textbox automatically.

## Problem

That pane's composer used a plain SwiftUI `TextEditor` with `.focused($draftComposerFocused)`, which does not reliably accept programmatic focus on macOS when the field has just appeared. So opening the composer left it unfocused, requiring an extra click before typing.

## Fix

Reuse the `NSViewRepresentable` composer (`ReviewDraftComposerTextEditor`) already used by the review and commit surfaces, which makes the backing text view first responder on appear via `window.makeFirstResponder`. The composer is now focused automatically as soon as it opens.

- `DiffReviewFileSection.swift`: made `ReviewDraftComposerTextEditor` module-internal so it can be shared.
- `DiffTabView.swift`: replaced the plain `TextEditor` with the shared composer, wired to save/cancel, and focus it on appear.

As a consistency bonus, this brings the pane's composer keyboard shortcuts (Cmd-Return to save, Esc to cancel) in line with the other surfaces.

## Verification

Full `xcodebuild` for macOS succeeds.

### DIFF
--- a/Alas/Sources/Center/DiffReview/DiffReviewFileSection.swift
+++ b/Alas/Sources/Center/DiffReview/DiffReviewFileSection.swift
@@ -792,7 +792,7 @@ enum ReviewDraftComposerKeyboardAction: Equatable {
     }
 }
 
-private struct ReviewDraftComposerTextEditor: NSViewRepresentable {
+struct ReviewDraftComposerTextEditor: NSViewRepresentable {
     @Binding var text: String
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #653, opened 2026-07-09
fix(acp): codex session restore

mrmans0n:
## Summary
Fix Codex ACP session restore by accepting successful `session/load` responses that omit a top-level `sessionId`, which matches the Codex ACP behavior observed in practice. This prevents restored Codex ACP panes from falling back to lossy transcript recovery when normal restore already succeeded.

## Changes
- Fill a missing `session/load` result id from the requested remote session id.
- Keep `session/new` strict so new sessions still require a returned id.
- Add a regression test for a Codex-style load response with nested capabilities and no top-level `sessionId`.

## Verification
- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/ACPSessionManagerAttachRestoreTests test`
- Started full `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test`; interrupted after Xcode stalled in test runner cleanup/worker materialization after several suites had already run.

### DIFF
--- a/Alas/Sources/ACP/Protocol/ACPConnection.swift
+++ b/Alas/Sources/ACP/Protocol/ACPConnection.swift
@@ -34,7 +34,14 @@ final class ACPConnection: @unchecked Sendable {
         let req = ACPRequest(method: "session/new",
                              params: ACPSessionNewParams(cwd: cwd, mcpServers: []))
         let resp = try await client.send(req)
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #662, opened 2026-07-10
feat(stashes): collapse the stashes section by default

mrmans0n:
Stashes are a secondary concern in the changes pane, so start the
section collapsed rather than expanded to keep the default view
focused on the working-tree changes.


mrmans0n:
@codex review

PR #664, opened 2026-07-10
fix(acp): filter model picker by visible names

mrmans0n:
## Summary
- Make the model picker search match visible model names only
- Keep identifier and description search available for the shared select chip elsewhere
- Add a regression test for hidden model identifiers that previously kept unrelated rows visible

## Validation
- xcodegen
- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/ACPSelectChipTests

Note: the full local test suite was attempted, but Xcode was interrupted while finalizing runner output after several minutes without producing a failure summary.


mrmans0n:
@codex review

PR #715, opened 2026-07-10
perf(diff): precompute display digests to keep review body O(1)

### LINKED ISSUE #676: perf(diff): DiffReviewFileSection body is O(file-size) on every SwiftUI update
**Severity: high**

Two independent O(all-rows) computations run per body evaluation of every realized file section in the review stream:

- `displayGroupSignature` (`Alas/Sources/Center/DiffReview/DiffReviewFileSection.swift:736-748`) builds a joined string over every row; it feeds `.onChange(of: draftCommentDisplaySignature)` (line 112) and `.onChange(of: contextStateSignature)` (line 118) — `.onChange` re-evaluates its value on **every** body pass, so this string is built **twice per update**.
- `renderContext` (line 90 / 345-378) constructs `DiffReviewRenderContextKey`, whose `DisplayModelSignature` maps every group/row/line into signature structs using reflection — `String(describing: row.kind)` per row (`DiffReviewRenderContext.swift:237-254`) — then hashes it all for the cache lookup.

Body updates fire constantly during scrolling: the scroll spy writes `selectedFileID` (`DiffReviewSurface.swift:254-259`), and hover state (`hoveredDraftCommentID`, `hoveredInlineFeedbackID`, `DiffReviewFileSection.swift:84-85`) invalidates the section. Each invalidation costs O(lines) in string building, reflection, and hashing on main — tens of thousands of allocations per scroll tick for a 5k-line file.

**Fix direction:** compute signatures once at model-build time (store a digest on `DiffDisplayModel`/`DiffDisplayGroup`); replace `String(describing: kind)` with the `Hashable` enum itself; key the render-context cache on cheap identity + a precomputed content hash.

Same shape exists for single-file diff tabs: `DiffTabRenderContextKey` (`DiffTabView.swift:543-552`, `DiffTabRenderContext.swift:187-199, 232-243`) — tracked separately.

From the 2026-07-10 performance audit (DIFF-H2).

mrmans0n:
## Summary

`DiffReviewFileSection.body` ran two O(all-rows) computations on **every** SwiftUI update, and body updates fire constantly during review (scroll spy writes `selectedFileID`, hover state invalidates the section). Each invalidation cost O(lines) in string building, reflection, and hashing on the main actor.

This moves the per-row work to **model-build time** so body passes read precomputed scalars in O(1):

- `DiffDisplayModel`/`DiffDisplayGroup` now carry `contentHash` (full fidelity — keys the render-context cache) and `structuralHash` (coarse — drives the `.onChange` state resets), computed once in their initializers. `DiffDisplayGroup` also carries precomputed old/new side extents.
- The render-context key stores `contentHash` instead of rebuilding `DisplayModelSignature` (which mapped every group/row/line into signature structs with `String(describing:)` reflection per row before hashing). The dead signature structs are deleted.
- `BoundaryRange` reads the precomputed extents instead of an O(hunk-lines) `reduce` that ran per body pass whenever a context provider was present.
- The section's draft/context `.onChange` signatures become O(1) reads of `structuralHash` via small `Equatable` structs, replacing a joined string that was rebuilt twice per pass.
- `DiffDisplayRow.Kind` is now `Hashable`, replacing `String(describing:)`.

`DiffTabView`'s parallel `DiffTabRenderContextKey` has the same shape and is tracked separately by the issue.

## Testing

- `xcodebuild build` — succeeds.
- New `DiffDisplaySignatureTests` covers `contentHash` vs `structuralHash` semantics (text edits change content but not structure; line-number/collapse changes move both), precomputed side extents, and that the render key still distinguishes model content while treating identical models as equal.
- All diff/review/render suites pass (`DiffReviewSurfaceTests`, `DiffContextExpansionTests`, `DiffReviewModelsTests`, `DiffReviewFileSectionEqualityTests`, `DiffPaneViewTests`), including the existing test that verifies expanded-context text changes still invalidate the cache through the new content-hash key.

Closes #676

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -683,6 +683,7 @@
 		9A3EB3ACA3AB366431C904EB /* WorktreeService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6C4E238E48247A6CDD5565D2 /* WorktreeService.swift */; };
 		9AB33F6B645EBA3F92A4E40F /* ACPSessionUpdate.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD1F34E025D235A5CECCE36E /* ACPSessionUpdate.swift */; };
 		9AC7E08307DA010181E2DE58 /* ACPFirstRunConnectingPolicy.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8B4BB6C10F3718F7D0A50BAE /* ACPFirstRunConnectingPolicy.swift */; };
+		9B00878F781FADB428684B2D /* DiffDisplaySignatureTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 283613E170E25F69521F64ED /* DiffDisplaySignatureTests.swift */; };
 		9B2B055195430A5A40BC7659 /* TreeSitterJSON in Frameworks */ = {isa = PBXBuildFile; productRef = 5130D28B6980AF12D2256A6D /* TreeSitterJSON */; };
 		9B51DB62236B865AB86AA0F1 /* TreeSitterTypeScript in Frameworks */ = {isa = PBXBuildFile; productRef = 1F70A188A5F7992AC8C8717E /* TreeSitterTypeScript */; };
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #717, opened 2026-07-10
fix(acp): transcript render scaling

### LINKED ISSUE #666: perf(acp): transcript is a non-lazy VStack and the reveal window only ever grows
**Severity: high**

`ACPMessageList` renders the transcript as `ScrollView { VStack { ForEach(visibleRows, id: \.message.stableId) } }` (`Alas/Sources/ACP/UI/ACPMessageList.swift:108-131`). Not lazy: every revealed row is fully instantiated and kept alive — each agent/user paragraph is an `NSTextView` (`ACPMarkdownInlineTextView`), each row carries a `GeometryReader` background (line 130) and an always-mounted AppKit `Menu` in `ACPMessageGutter` (`ACPMessageGutter.swift:60-84`).

The `visibleHead` window starts at the last 30 rows (`ACPTranscript.tailWindow`, `ACPTranscript.swift:40`), but `stepHeadBack()` (`ACPTranscript.swift:125-127`) only decrements — nothing re-advances the head as you scroll back down. Scrolling to the top of a 1,000-message transcript leaves 1,000 live rows in a non-lazy stack; layout/diffing cost then scales with everything revealed and scrolling gets progressively worse the further up you've been.

**Fix direction:** `LazyVStack` inside the `ScrollView` (keeps `scrollTargetLayout`/visibility APIs working), or actively re-shrink the window; drop the per-row `GeometryReader` on macOS 15 (see the companion issue about the preference pipeline).

From the 2026-07-10 performance audit (ACP-H1).

mrmans0n:
## Summary

Fixes #666.

- Render ACP transcript rows in a `LazyVStack` so previously revealed long histories do not eagerly instantiate every message row.
- Keep row frame reporting on the macOS 14 fallback path only; macOS 15+ continues using scroll target visibility without per-row `GeometryReader` preferences.
- Recompact the transcript window back to the tail while tail-follow is active, and cover tail-window cache eviction with a focused test.

## Validation

- `xcodegen`
- `ALAS_FFF_TARGET_ARCH=arm64 rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/ACPTranscriptMarkdownCacheEvictionTests test`

Note: a full local `xcodebuild ... test` run was started before publishing but was interrupted after staying silent for several minutes; this PR relies on CI for the exhaustive pass.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPMessageList.swift
+++ b/Alas/Sources/ACP/UI/ACPMessageList.swift
@@ -40,7 +40,7 @@ struct ACPMessageList: View {
     @State private var restoredRememberedAnchor: String?
     @State private var pendingTailScrollTask: Task<Void, Never>?
 
-    /// Height of an invisible spacer at the tail of the VStack. The
+    /// Height of an invisible spacer at the tail of the transcript stack. The
     /// composer pill plus its outer padding occupies roughly this much
     /// vertical space, so by scrolling THAT element to the viewport
     /// bottom we guarantee the streaming caret / last message sits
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #718, opened 2026-07-10
perf(acp): gate row frame tracking to legacy scroll

### LINKED ISSUE #669: perf(acp): per-row GeometryReader preference pipeline still mounted on macOS 15
**Severity: medium**

Every transcript row carries a `GeometryReader` publishing `[String: CGRect]` (`Alas/Sources/ACP/UI/ACPMessageList.swift:130, 440-447, 810-815`); the preference `reduce` merges dictionaries O(rows) per layout pass; `handleRowFramePreference` (407-425) rebuilds `visibleMessageLookup` (O(n)) and runs a `topVisibleAnchorID` min-scan (542-553).

Comments say this pipeline is dead during scroll on macOS 15+ (superseded by `onScrollTargetVisibilityChange`), yet it still fires on every layout/streaming change; on macOS 14 it fires per scroll frame.

**Fix direction:** gate `rowFrameReporter` + `onPreferenceChange` behind `if #unavailable(macOS 15)`.

From the 2026-07-10 performance audit (ACP-M1).

mrmans0n:
## Summary

Fixes #669 by keeping the ACP transcript row-frame preference pipeline on the macOS 14 fallback path only.

- avoids mounting per-row `GeometryReader` frame reporters on macOS 15+
- avoids attaching the row-frame `onPreferenceChange` handler on macOS 15+
- keeps the existing legacy preference path for macOS 14
- adds focused coverage for the legacy-only tracking decision

## Root cause

`ACPMessageList` still installed the old row-frame `PreferenceKey` pipeline even on macOS 15+, where scroll tracking has moved to SwiftUI scroll geometry and scroll target visibility APIs. That left an O(rows) preference merge and anchor scan active during layout/streaming changes even though the path was no longer the source of scroll truth on modern macOS.

## Validation

- `rtk xcodegen`
- `ALAS_FFF_TARGET_ARCH=arm64 rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/ACPMessageListPaginationTests test`

Note: a full local `xcodebuild ... test` run was started before publishing, but stayed silent for an extended period and was interrupted rather than reported as green.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPMessageList.swift
+++ b/Alas/Sources/ACP/UI/ACPMessageList.swift
@@ -275,9 +275,9 @@ struct ACPMessageList: View {
                 .onChange(of: transcript.visibleHead) { _, _ in
                     restoreRememberedAnchorIfNeeded(proxy: proxy)
                 }
-                .onPreferenceChange(ACPRowFramesPreferenceKey.self) { frames in
+                .modifier(ACPRowFramePreferenceTracking { frames in
                     handleRowFramePreference(frames, proxy: proxy)
-                }
+                })
             }
         }
         .background(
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #724, opened 2026-07-11
fix(acp): clear sent composer drafts on restore

mrmans0n:
## Summary

Fixes ACP composer restore so a message that was already sent before quitting Alas does not come back pre-filled after relaunch.

## Root cause

The submit path intentionally persisted the submitted composer draft as crash recovery while the prompt RPC was in flight. If Alas exited after the prompt had already progressed and transcript output was persisted, but before the composer success callback deleted that recovery row, hydration could treat the old submitted text as an unsent draft. Timestamp-only cleanup was not precise enough because draft writes and transcript rows use second-resolution timestamps, and repeated prompts can have the same text as older transcript turns.

## Changes

- Keep the suspended composer draft while prompt delivery is still in flight, so a locally recorded but undelivered prompt remains retryable after restore.
- Add a `submitted_recovery` marker to composer draft persistence so restore can distinguish submitted recovery rows from normal typed drafts, including same-second repeated text.
- Store the latest transcript sequence alongside submitted recovery drafts and only match them against user messages recorded after that boundary, so older duplicate turns do not delete a newly submitted recovery draft.
- Normalize image-chip drafts before matching persisted user prompts, so sent image prompts with trailing chip whitespace are cleared once agent progress exists.
- During hydration, discard submitted recovery rows that match a persisted user prompt only after later agent-side progress exists, while preserving genuinely typed drafts and recorded queued prompts.
- Make restore-time stale-draft cleanup conditional on the exact row that was inspected, so a newer draft written during hydration is not deleted.
- Add focused regressions for restore-time cleanup, undelivered prompt preservation, same-second submitted recovery cleanup, same-second typed draft preservation, duplicate prompt boundaries, image prompt normalization, recorded queued prompt preservation, schema migration, and conditional draft deletion.

## Validation

```bash
rtk env ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/ACPSessionManagerHydrationTests -only-testing:AlasTests/ACPSessionRunnerTests -only-testing:AlasTests/ACPSessionStoreCRUDTests -only-testing:AlasTests/ACPSessionStoreSchemaTests -only-testing:AlasTests/ACPSessionLeaseTests -only-testing:AlasTests/ACPSessionHydratorTests -only-testing:AlasTests/ACPSessionStoreQueueTests -only-testing:AlasTests/ACPSessionStoreDedupTests -only-testing:AlasTests/ACPComposerDraftTests test
```

Passed: 137 tests in 9 suites.

### DIFF
--- a/Alas/Sources/ACP/Session/ACPComposerDraft.swift
+++ b/Alas/Sources/ACP/Session/ACPComposerDraft.swift
@@ -87,6 +87,68 @@ struct ACPComposerDraft: Codable, Equatable {
 }
 
 extension ACPComposerDraft {
+    func matchesPersistedUserPrompt(text: String, attachments: [ACPMessage.Attachment]) -> Bool {
+        normalizedPromptBlocks == Self.contentBlocks(text: text, attachments: attachments)
+    }
+
+    func matchesRecordedQueuedPrompt(in queue: [QueuedPrompt]) -> Bool {
+        queue.contains { item in
+            item.transcriptRecorded && self == item.restorableDraft
+        }
+    }
+
+    private var normalizedPromptBlocks: [ACPContentBlock] {
+        var text = ""
+        var attachments: [ACPMessage.Attachment] = []
+        for segment in segments {
+            switch segment {
+            case .text(let value):
+                text += value
+            case .mention(let displayName, let uri):
+                text += "@\(displayName) "
+                attachments.append(.init(uri: uri, name: displayName))
+            case .image(let uri, let mimeType):
+                attachments.append(.init(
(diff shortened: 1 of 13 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #725, opened 2026-07-11
chore(treesitter): Update tree-sitter Swift grammar

mrmans0n:
## Summary

- Bump the `TreeSitterSwift` generated-files revision from `0.7.2-with-generated-files` to `0.7.3-with-generated-files`.
- Regenerate `Alas.xcodeproj` with `xcodegen` and update `Package.resolved` to the same revision.

## Audit notes

- Vendored tree-sitter grammars (`CSS`, `HCL`, `JavaScript`, `Lua`, `Python`, `YAML`) are already at their latest upstream semver tags.
- Other direct tree-sitter grammar package pins are already current.
- `tree-sitter` runtime has a newer `v0.26.10`, but `swift-tree-sitter 0.25.0` constrains it to `.upToNextMinor(from: "0.25.0")`, so the runtime stays at `0.25.10`.

## Verification

- `rtk xcodegen`
- `rtk xcodebuild -resolvePackageDependencies -project Alas.xcodeproj -scheme Alas`
- `ALAS_FFF_TARGET_ARCH=arm64 rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/TreeSitterHighlighterTests test`

Full `xcodebuild ... test` was started locally, but `xcodebuild` hung silently with no active test child after the focused tree-sitter suite and full app build passed.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -5976,7 +5976,7 @@
 			repositoryURL = "https://github.com/alex-pinkus/tree-sitter-swift";
 			requirement = {
 				kind = revision;
-				revision = 7b7909f2f6b9414be0958275f4c8e5d69c3bca43;
+				revision = 31d17fe7e818a2048c808b5c6fdc2dc792f4f5b5;
 			};
 		};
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #729, opened 2026-07-11
fix(harness): nudge churn

### LINKED ISSUE #698: perf(ui): HarnessDetector churns observable state at 1 Hz; nudge banner does disk I/O in body
**Severity: medium**

`HarnessDetector` (`Alas/Sources/Harness/HarnessDetector.swift:24, 35-41`): a 1-second timer calls `onUpdate(sid, kind)` for every registered terminal session every tick. `HarnessService.recordHarnessDetection` (`HarnessService.swift:61-87`) assigns `harnessBySession[sid]` / `activeHarnessBySession[sid]` unconditionally — same value still fires the `@Observable` mutation hook, and the nil path calls `removeValue` every tick (dictionary `_modify` fires even when the key is absent). Net: `activeHarnessBySession` invalidates its observers ~1 Hz whenever any terminal exists.

Consumer: `AgentHookInstallNudgeBanner` (`AgentHookInstallNudgeBanner.swift:13-27`) — the computed `nudge` runs in `body`; when a hook-capable agent is detected it constructs `AgentInstallerRegistry()` and calls `installState()` → `JSONHookSettingsFile.load` (file read + JSONSerialization, e.g. `ClaudeInstaller.swift:16-35`) — synchronous disk I/O in `body`, re-run every ~1 s while an agent runs. The comment claims memoization but none exists.

**Fix direction:** equality-guard the writes in `recordHarnessDetection` (mirror the documented pattern in `RightPaneState.refresh`); cache the nudge result keyed on (detected harnesses, dismissed set) in `@State`/task instead of computing `installState()` in body.

Related (medium): `HarnessService.activityBySession` single dict fans every agent state-transition socket event across all project groups + tab bar (`HarnessService.swift:10, 89-178`; `SidebarView.swift:51-60` also flat-maps tabs per worktree per render). Acceptable at current scale; keyed access if the sidebar grows.

From the 2026-07-10 performance audit (UI-M1, UI-M3).

mrmans0n:
Fixes #698.

Summary:
- Skips unchanged HarnessService detection writes so polling does not invalidate observers.
- Resolves the hook install nudge from a cached task key instead of running installer state checks in SwiftUI body.
- Adds observation tests for no-op detection updates.

Validation:
- ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination platform=macOS -only-testing:AlasTests/HarnessServiceTests test
- ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination platform=macOS -quiet build
- ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination platform=macOS test was interrupted after unrelated existing failures/hang in ACPSession long image URL asset tests and DiffPaneViewTests syntax highlighting.

### DIFF
--- a/Alas/Sources/Center/AgentHookInstallNudgeBanner.swift
+++ b/Alas/Sources/Center/AgentHookInstallNudgeBanner.swift
@@ -77,6 +77,8 @@ struct AgentHookInstallNudgeBanner: View {
                     installStatus = nil
                     resolvedNudge = nil
                 }
+                let key = await MainActor.run { nudgeKey }
+                await refreshNudge(for: key)
             } catch {
                 await MainActor.run {
                     installStatus = "Error: \(error.localizedDescription)"
@@ -95,6 +97,7 @@ struct AgentHookInstallNudgeBanner: View {
         }
     }
 
+    @MainActor
     private func refreshNudge(for key: HookInstallNudgeKey) async {
         guard !key.detectedHarnesses.isEmpty else {
             resolvedNudgeKey = key
--- a/Alas/Sources/Center/AgentHookInstallNudgeBanner.swift
+++ b/Alas/Sources/Center/AgentHookInstallNudgeBanner.swift
@@ -33,6 +33,10 @@ struct AgentHookInstallNudgeBanner: View {
         .task(id: currentNudgeKey) {
             await refreshNudge(for: currentNudgeKey)
         }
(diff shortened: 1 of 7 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #741, opened 2026-07-12
perf(acp): trim markdown transcript rendering churn

### LINKED ISSUE #674: perf(acp): low-severity transcript rendering cleanups
**Severity: low** — grab bag from the 2026-07-10 performance audit.

- `ACPMarkdownText.inlineMarkdown` NSCache (`ACPMarkdownText.swift:104-135`): every streamed prefix of the tail paragraph is inserted as a distinct key into the 512-limit cache — churn evicts warm entries for stable blocks.
- `MarkdownImageLoader` instance per `ACPMarkdownInlineTextView.Coordinator` (`ACPMarkdownInlineTextView.swift:58`) — remote image cache is not shared across rows; the same image is fetched N times.
- `ACPMarkdownInlineTextView.updateNSView` (lines 30-39) has no internal early-out; it relies entirely on SwiftUI value diffing, which the fresh-String/closure issues partially defeat.
- `ACPMarkdownText` `ForEach(... id: \.offset)` (line 22): offset identity is fine for append-only promotion but re-keys everything if an early block changes; acceptable today, noting as a trap.

mrmans0n:
## Summary
- avoid populating the inline markdown cache from streaming tail blocks while keeping stable blocks memoized
- skip redundant `NSTextView` updates when ACP inline markdown render inputs are unchanged
- share and coalesce remote markdown image loads across inline rows

Fixes #674.

## Verification
- `xcodegen`
- `ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build`
- `ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/MarkdownImageLoaderTests -only-testing:AlasTests/ACPMarkdownTextBareURLTests test`

Note: the full local test command was started but remained active without a terminal result after an extended wait, so I interrupted it and left broad coverage to CI.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPMarkdownInlineRenderer.swift
+++ b/Alas/Sources/ACP/UI/ACPMarkdownInlineRenderer.swift
@@ -91,10 +91,13 @@ enum ACPMarkdownInlineRenderer {
         source: String,
         theme: Theme,
         typography: ACPChatTypography,
-        role: ACPMarkdownInlineRole
+        role: ACPMarkdownInlineRole,
+        memoizeInlineMarkdown: Bool = true
     ) -> NSMutableAttributedString {
         let plan = makePlan(source)
-        let attributed = NSMutableAttributedString(ACPMarkdownText.inlineMarkdown(plan.markdownSource))
+        let attributed = NSMutableAttributedString(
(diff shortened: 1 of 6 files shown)


mrmans0n:
@codex review

PR #745, opened 2026-07-12
fix(shortcuts): ignore unsupported menu shortcuts

mrmans0n:
## Summary

- Treat unsupported persisted shortcut keys as unbound instead of converting them to Space.
- Reuse the same supported-key predicate when recording shortcuts.
- Reject unsupported shortcut assignments before they reach persisted config.

## Root Cause

ShortcutBinding converted any unknown multi-character key into a Space key equivalent. A stale or malformed shortcut override could therefore feed bogus shortcut metadata into SwiftUI's AppKit menu bridge while the main menu was being rebuilt.

## Validation

- `xcodegen`
- `ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' -quiet build`
- `ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' -only-testing:AlasTests/ShortcutBindingTests -only-testing:AlasTests/ShortcutResolverTests test`

Full local `xcodebuild ... test` was attempted but interrupted after it wedged for several minutes following unrelated existing failures/logged exceptions in ACP/DiffPane tests.

### DIFF
--- a/Alas/Sources/Shortcuts/ShortcutBinding.swift
+++ b/Alas/Sources/Shortcuts/ShortcutBinding.swift
@@ -41,11 +41,21 @@ struct ShortcutBinding: Codable, Equatable, Hashable, Sendable {
         }
     }
 
-    func asKeyboardShortcut() -> KeyboardShortcut {
-        KeyboardShortcut(keyEquivalent, modifiers: eventModifiers)
+    var hasSupportedKey: Bool {
+        keyEquivalent != nil
     }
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #746, opened 2026-07-12
fix(acp): stabilize task sidebar placement

mrmans0n:
## Summary

- measure task sidebar visibility from the full ACP pane
- keep chat content sizing tied to the measured chat column
- prevent sidebar insertion/removal from feeding its reduced width back into placement

## Root cause

The visibility decision was attached inside the transcript/sidebar layout subtree. Recent transcript layout changes could make that measurement reflect the width after the 320-point task sidebar was inserted, causing the visibility reducer to hide it again. Removing it restored the wider measurement and repeated the cycle between the toolbar card and right sidebar.

## Verification

- `xcodegen`
- focused `ACPPlanSidebarVisibilityTests`, `ACPChatLayoutTests`, and `ACPSessionManagerPlanSidebarLayoutTests`
- quiet macOS build

The broader local test run passed the ACP suites, then stalled after an unrelated AppKit `NSRangeException` in the diff/editor test surface.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPTabView.swift
+++ b/Alas/Sources/ACP/UI/ACPTabView.swift
@@ -107,32 +107,52 @@ private struct ACPSessionView: View {
     }
 
     var body: some View {
-        VStack(spacing: 0) {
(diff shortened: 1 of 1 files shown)


mrmans0n:
@codex review

PR #747, opened 2026-07-12
fix(acp): avoid ACPMessage copies in transcript rows

mrmans0n:
## Summary

- replace ACP transcript visible row data with a lightweight index/stable-id descriptor
- fetch the full `ACPMessage` only inside row rendering, outside SwiftUI's `ForEach` identity path
- add focused coverage for the visible-row projection that skips plan rows while preserving transcript indices

## Root Cause

`ACPMessageList` fed `(index, stableId, message: ACPMessage)` tuples directly into `ForEach`. Even with `id: \.stableId`, SwiftUI still copied and destroyed the heavy enum payload while generating lazy-stack row identity, which showed up in samples under `ForEach.IDGenerator.makeID`.

## Validation

- `rtk xcodegen`
- `ALAS_FFF_TARGET_ARCH=arm64 rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/ACPMessageListPaginationTests -only-testing:AlasTests/ACPMessageStableIdTests test`

### DIFF
--- a/Alas/Sources/ACP/UI/ACPMessageList.swift
+++ b/Alas/Sources/ACP/UI/ACPMessageList.swift
@@ -59,13 +59,12 @@ struct ACPMessageList: View {
     /// reset to `max(0, count - tailWindow)` after hydration); the
     /// filter drops `.plan` entries because the toolbar pill renders
     /// the current turn's plan instead of an inline card.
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #748, opened 2026-07-12
fix(acp): stabilize transcript tail scrolling

mrmans0n:
## Summary

Fixes a transient ACP transcript blanking risk during live streaming by reducing avoidable lazy-stack invalidation and preventing tail-scroll correction from being starved by rapid updates.

## Details

- Avoid republishing `visibleHead` when `resetWindowToTail()` computes the same head.
- Coalesce pending tail-scroll tasks so fast streaming updates do not repeatedly cancel the scroll-to-tail correction before it can run.
- Add focused regression coverage for no-op transcript window resets and tail-scroll scheduling coalescing.

## Validation

- `rtk git submodule update --init ThirdParty/zmx ThirdParty/fff`
- `ALAS_FFF_TARGET_ARCH=arm64 rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/ACPTranscriptWindowTests -only-testing:AlasTests/ACPMessageListPaginationTests test`
- `rtk xcodegen`

A full local `xcodebuild ... test` was started before publishing but stayed silent for several minutes and was interrupted without failure output; CI should provide the broad suite signal.

### DIFF
--- a/Alas/Sources/ACP/Session/ACPTranscript.swift
+++ b/Alas/Sources/ACP/Session/ACPTranscript.swift
@@ -144,6 +144,7 @@ final class ACPTranscript: ObservableObject {
     /// `visibleHead` directly when advancing the window.
     func setVisibleHead(_ newHead: Int) {
         let clamped = max(0, min(newHead, messages.count))
+        guard clamped != visibleHead else { return }
         guard clamped > visibleHead else {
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

PR #761, opened 2026-07-13
fix(anr): Cap Mason extension install options

mrmans0n:
## Summary

- cap the install nudge menu to `MasonSnapshot.maxResults` after recipe, installer, availability, and duplicate filtering
- keep Mason extension lookup ranked and complete so later installable matches are not hidden by earlier unavailable ones
- add regression coverage for late installable matches, duplicate package IDs, and capped final menu options

## Root Cause

The install nudge menu picker is fed by Mason extension lookup results. The menu needed a bounded, stable option set, but capping or deduplicating before installability filtering can hide a later package that is actually installable. The corrected flow scans the ranked Mason matches, filters for installable options, deduplicates only surviving options by `masonId`, and stops after 25 final menu options.

## Validation

- `rtk xcodegen`
- `rtk env ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build`
- `rtk env ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/MasonSnapshotTests -only-testing:AlasTests/InstallNudgeResolverTests test`

Full local suite was also run before publish and reached unrelated ACP queue/recovery and SSH integration failures already outside this change scope.

### DIFF
--- a/Alas/Sources/Code/Editor/InstallNudgeResolver.swift
+++ b/Alas/Sources/Code/Editor/InstallNudgeResolver.swift
@@ -103,19 +103,24 @@ struct InstallNudgeResolver {
         guard !dismissedInstallNudges.contains(dismissalKey) else { return nil }
         guard !registry.disabledUserDefinedEntryClaims(fileExtension: ext) else { return nil }
 
-        let options = masonSnapshot.packages(forFileExtension: ext).compactMap { package -> InstallNudgeOption? in
-            guard !package.recipes.isEmpty else { return nil }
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #767, opened 2026-07-13
fix(git): prevent worktree deletion launch crash

mrmans0n:
## Summary

- run local Git commands with `git -C` instead of setting `Process.currentDirectoryURL`
- preserve launch errors for worktrees that are already missing while making post-preflight deletion a normal Git failure
- cover local invocation shape and the delete-before-launch race

## Testing

- `xcodegen`
- focused `GitInvocationTests`, `ProcessGitTests`, and `RightPaneStateBaseBranchTests`
- `ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination "platform=macOS" -quiet build`
- full local suite: 5,233 passed; 13 unrelated failures from opt-in SSH integration checks and existing ACP concurrency/process tests

### DIFF
--- a/Alas/Sources/Git/Process+Git.swift
+++ b/Alas/Sources/Git/Process+Git.swift
@@ -191,10 +191,14 @@ extension Process {
         stdin: String? = nil,
         timeout: TimeInterval = Process.defaultTimeout
     ) async throws -> ProcessResult {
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

PR #769, opened 2026-07-13
feat(ssh): add interactive connection assistant

mrmans0n:
## Summary
- replace headless first-contact SSH validation with a batch fast path and an embedded interactive terminal
- verify the shared ControlMaster connection before retrying project creation
- distinguish recoverable authentication failures from servers that reject required non-interactive channels

## Why
The existing interactive preflight only removed BatchMode while still running through Foundation.Process without a TTY. Password, host-key, keyboard-interactive, and hardware-key prompts could not render, so users had to copy a generated SSH command into another terminal.

## Validation
- `xcodegen`
- focused `RemoteRepoValidatorTests` and `SSHCommandTests`
- macOS quiet build
- full suite: 5,238 passed; four unrelated ACP queue persistence tests remain order-dependent and pass in isolation

### DIFF
--- a/Alas/Sources/Dialogs/NewProjectDialog.swift
+++ b/Alas/Sources/Dialogs/NewProjectDialog.swift
@@ -26,6 +26,41 @@ private enum ProjectDialogMode {
     case edit(ProjectConfig)
 }
 
+private struct SSHConnectionIssue {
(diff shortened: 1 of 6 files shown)


mrmans0n:
@codex review

PR #770, opened 2026-07-13
feat(headless): low-latency fs and search ops

### LINKED ISSUE #753: Remote helper: low-latency fs and search ops over the persistent channel
Part of #749. Depends on #750, #751.

## Problem

Every `RemoteFileAccess.read/write/mtime`, untracked line count, lazy directory listing, and content search is a discrete ssh exec (#738). ControlMaster makes each ~1 RTT plus process spawn overhead on both ends; interactive flows that fan out (editor open + ACP fs traffic + tool-card images) feel it, especially on high-latency links.

## Scope

1. Route `RemoteFileAccess` through the helper channel when available: `fs/read`, `fs/write` (atomic, permission-preserving, `expectedMtime` conflict gate server-side — strictly better than the client-side stat-then-write race from #738), batched `fs/stat`.
2. Consumers come along for free — editor load/save, `ACPRemoteFileServer` (fs/read+fs/write for agents), `RemoteImageCache`, `DiffPaneLSP` reads — keep `RemoteFileAccess`'s public API so call sites don't change; the helper-vs-exec choice lives inside.
3. `RemoteFileStats` (untracked `wc -l` batches, `ls -1Ap` listings) → helper equivalents (one request, no 200-file argv cap).
4. Content search: helper streams `rg --json` (or its own matcher) over the channel with incremental results and server-side cancellation — replaces the per-search ssh exec; `git grep` fallback stays for helper-less hosts.
5. Measure: add a simple per-op timing log (debug-gated) so the win is quantifiable before/after.

## Acceptance

- Editor open/save and agent fs round trips over a ~100ms-RTT link are visibly snappier with the helper (target: 1 RTT + no ssh process spawn per op).
- Server-side mtime gate: concurrent external write between load and save yields a conflict, deterministically (no race window).
- Helper-less hosts: unchanged #738 behavior, including the wc/ls fallbacks.

## Non-goals

Routing git plumbing (`Process.git`) through the helper — git-over-exec is fine and battle-tested; revisit only if profiling says otherwise.

mrmans0n:
## Summary

- route remote reads, writes, stats, line counts, and directory listings through the persistent helper channel when available
- move the editor save conflict check into the helper's atomic write operation while preserving the existing SSH exec fallback
- stream `rg --json` search events over the helper channel with server-side cancellation and retain `git grep` on helper-less hosts
- add binary-safe reads, debug operation timing, protocol documentation, and focused Swift/Rust coverage

## Why

Remote file and search operations previously launched a separate SSH exec for every request. Even with ControlMaster, interactive workflows paid a round trip and remote process spawn per operation. The client-side stat-then-write save gate also left a race window before the write.

The persistent helper now owns these operations under registered roots. Older or unavailable helpers fall back to the existing exec implementations, so helper-less hosts keep the current behavior.

Closes #753.

## Validation

- `xcodegen`
- `ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build`
- focused `RemoteHelperClientTests`, `RemoteFileAccessTests`, `RemoteFileStatsTests`, and `RemoteContentSearchTests` (43 tests passed)
- full non-integration suite with `-skip-testing:AlasTests/SSHIntegrationTests` (passed; 5,262 tests passed in the preceding full run)
- `cargo test --manifest-path AlasHelper/Cargo.toml` (23 tests passed)
- `cargo fmt --manifest-path AlasHelper/Cargo.toml`
- `git diff --check`

`SSHIntegrationTests` could not run because localhost SSH is unavailable on the development machine (`port 22: Connection refused`). The ungated full run passed 5,262 tests and failed only the eight tests that require `ALAS_SSH_INTEGRATION=1` plus key-authenticated localhost SSH.

### DIFF
--- a/Alas/Sources/Code/Editor/EditorBuffer.swift
+++ b/Alas/Sources/Code/Editor/EditorBuffer.swift
@@ -1025,38 +1025,26 @@ final class EditorBuffer {
             try await verifyRemoteFileContained(host: host, path: path)
             priorOriginalText = originalText
             let baseline = originalMtime
-            let remoteMtime = try await RemoteFileAccess.mtime(host: host, path: path)
             let shouldOverwriteConflict = remoteOverwriteAfterConflict
-            if !shouldOverwriteConflict {
-                switch RemoteSaveGate.decision(
-                    originalMtime: baseline,
-                    remoteMtime: remoteMtime
-                ) {
-                case .conflict:
-                    guard remoteSaveGeneration == generation else { return }
-                    rollbackRemoteSave(to: priorOriginalText, conflict: .changedOnDisk)
-                    throw SaveError.remoteSaveConflict
-                case .targetDeleted:
-                    guard remoteSaveGeneration == generation else { return }
-                    rollbackRemoteSave(to: priorOriginalText, conflict: .deletedOnDisk)
-                    throw SaveError.remoteSaveConflict
-                case .proceed:
-                    break
-                }
-                if RemoteSaveGate.requiresContentCheck(originalMtime: baseline, remoteMtime: remoteMtime) {
-                    guard try await remoteContentsStillMatchBaseline(host: host, path: path, baseline: priorOriginalText) else {
-                        guard remoteSaveGeneration == generation else { return }
-                        rollbackRemoteSave(to: priorOriginalText, conflict: .changedOnDisk)
-                        throw SaveError.remoteSaveConflict
-                    }
-                }
+            let newMtime: Date
+            do {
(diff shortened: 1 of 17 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #780, opened 2026-07-14
fix(diff): focus review comment composers

mrmans0n:
## Summary

- Focus the shared draft review composer after its underlying `NSTextView` attaches to a window.
- Add explicit SwiftUI focus state for inline review reply and edit text boxes.
- Fix two post-rebase split-view drag width inference issues exposed by the current main branch/Xcode toolchain.

## Root cause

The draft composer could request first responder status before the AppKit text view had a window, so the request was dropped in panes that mount the composer asynchronously, including "Review current changes".

## Validation

- `xcodegen`

Full local `xcodebuild ... test` was run before the rebase; it executed 5,383 tests and only failed the SSH integration tests that require `ALAS_SSH_INTEGRATION=1`.


mrmans0n:
@codex review

PR #782, opened 2026-07-14
fix(diff): scroll diff to comment when its rail card is clicked

mrmans0n:
Clicking a comment in the draft review summary rail already fired a
scroll command, but the target card is nested inside its file section —
a lazy child of the review stream's LazyVStack. When the comment's file
wasn't currently realized (the reviewer was viewing another file),
scrollTo(cardID) silently no-oped and the diff never moved.

Mirror the file-rail scroll: for cross-file navigation, first land on the
file section (a direct lazy child SwiftUI reliably realizes), let layout
settle, then scroll to the nested card. When the file is already in view
the section is realized, so a single direct scroll reaches the card
without a detour through the file header. A programmatic-scroll token
suppresses the scrollspy during the move so it doesn't snap the
selection back to the topmost visible file.

### DIFF
--- a/Alas/Sources/Center/DiffReview/DiffReviewScrollSpy.swift
+++ b/Alas/Sources/Center/DiffReview/DiffReviewScrollSpy.swift
@@ -109,6 +109,14 @@ struct DiffReviewProgrammaticScrollController: Equatable {
         isSuppressing = false
     }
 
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #785, opened 2026-07-14
fix(menu): Avoid SwiftUI menu churn in message gutters

mrmans0n:
## Summary

- Replace the per-message SwiftUI `Menu` in `ACPMessageGutter` with a lightweight AppKit-backed button.
- Create the native `NSMenu` only when the gutter action is clicked.
- Keep copy behavior reading the latest streaming text before writing to the pasteboard.

## Root cause

The beachball sample in `/tmp/beachball2.log` showed the main thread inside SwiftUI/AppKit pop-up menu bridging while the transcript lazy stack was laying out rows. Every visible transcript row mounted a hidden SwiftUI `Menu`, so SwiftUI still rebuilt pop-up menu item accessibility and SF Symbol image state during view graph updates.

## Validation

- `ALAS_FFF_TARGET_ARCH=arm64 rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet test -only-testing:AlasTests/ACPMessageListPaginationTests -only-testing:AlasTests/ACPMessageStableIdTests -only-testing:AlasTests/ACPTranscriptWindowTests`
- `rtk git diff --check`

### DIFF
--- a/Alas/Sources/ACP/UI/ACPMessageGutter.swift
+++ b/Alas/Sources/ACP/UI/ACPMessageGutter.swift
@@ -14,21 +14,23 @@ import SwiftUI
 /// transcript container (`ACPMessageList`) so every row stays aligned. This
 /// wrapper only reveals the button and positions it into the reserved right
 /// lane via an offset.
-struct ACPMessageGutter<Content: View>: View {
-    enum CopySource {
-        case text(String)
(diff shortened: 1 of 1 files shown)


mrmans0n:
@codex review

PR #796, opened 2026-07-15
fix(search): Cmd-P search filtering reliability

mrmans0n:
## Summary

- move file ranking off the main actor so large repositories do not freeze query updates
- make the latest query authoritative for results, errors, selection, and loading state
- show the custom inline spinner in the search field while work is in progress
- preserve backend matches when fallback file enumeration fails

## Root cause

Fallback fuzzy ranking and sorting ran synchronously on the main actor. That delayed query changes and cancellation, while stale tasks could also clear loading state or publish partial results after a newer query had started.

## Validation

- `xcodegen`
- focused `SearchModelTests` and `FileSearchRankerTests`
- complete non-SSH test suite (5,400+ tests); SSH integration tests are opt-in via `ALAS_SSH_INTEGRATION=1`


mrmans0n:
@codex review

PR #801, opened 2026-07-15
fix(diff): align split diff feedback lanes

mrmans0n:
## Summary
- Split diff feedback comments into old/new-side lanes so side-specific comments stay visually attached to the correct pane.
- Reuse the same lane resolver across provider feedback, review threads, annotations, standalone diff drafts, and review-surface composers.
- Keep mixed/unknown changed-side feedback on the right/full-width fallback path depending on view context.

## Test Plan
- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` (local run completed 5,516 passing tests and failed only the opt-in `SSHIntegrationTests` gate because `ALAS_SSH_INTEGRATION=1` was not set)


mrmans0n:
@codex review

PR #805, opened 2026-07-15
fix(build): use rustup toolchain for fff

### LINKED ISSUE #804: build-fff.sh uses homebrew cargo, breaking universal (x86_64) builds when rustup isn't PATH-first
## Problem

`scripts/build-fff.sh` invokes the bare `cargo` resolved from `PATH`. Because every Rust build script prepends `/opt/homebrew/bin` to `PATH`:

```sh
export PATH="/opt/homebrew/bin:/usr/local/bin:${PATH}"
```

on a machine where Rust is installed via Homebrew (`/opt/homebrew/bin/cargo -> ../Cellar/rust/<ver>/bin/cargo`), that bare `cargo` is the Homebrew toolchain, **not** the rustup-managed one.

Homebrew's rust only ships the native (`aarch64-apple-darwin`) std. When `build-fff.sh` cross-compiles the `x86_64-apple-darwin` slice for a universal build (`target_arch=universal`, which happens whenever `CURRENT_ARCH=undefined_arch` during the Xcode run-script phase), the Homebrew `rustc` can't find `core` for x86_64 and the build fails:

```
error[E0463]: can't find crate for `core`
  = note: the `x86_64-apple-darwin` target may not be installed
error: could not compile `stable_deref_trait` (lib) ...
error: could not compile `memchr` (lib) ...
** BUILD FAILED **
```

The script's own guard does not help:

```sh
if command -v rustup >/dev/null 2>&1; then
    if ! rustup target list --installed | grep -qx "${cargo_target}"; then
        rustup target add "${cargo_target}"
    fi
fi
```

`rustup target add` adds the target to the rustup **default** toolchain, but the build then compiles with the Homebrew `cargo` that never consults rustup — so the added target is invisible to the actual compile.

## Why the other tools are fine

`scripts/build-alas-cli.sh` and `scripts/build-alas-helper.sh` resolve their compiler through rustup explicitly:

```sh
rustc_bin="$(${rustup_bin} which --toolchain "${rust_toolchain}" rustc)"
cargo_bin="$(${rustup_bin} which --toolchain "${rust_toolchain}" cargo)"
```

so they pick up the rustup toolchain (which has the x86_64 std) and build correctly. `build-fff.sh` is the outlier.

## Suggested fix

Make `build-fff.sh` resolve `cargo`/`rustc` through rustup the same way the CLI/helper scripts do (`rustup which --toolchain <ch> cargo`), and add the target to *that* toolchain rather than the default. Optionally support `ALAS_CARGO_BIN`/`ALAS_RUSTUP_BIN` overrides for parity with the other scripts.

## Impact

- CI is unaffected (its runners have the x86_64 std available to whichever cargo is on PATH).
- Local universal `xcodebuild` builds fail on Homebrew-rust machines.

## Workaround

Build the native slice only, which skips the x86_64 cross-compile:

```sh
ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas \
  -destination 'platform=macOS' ONLY_ACTIVE_ARCH=YES ARCHS=arm64 build
```

mrmans0n:
## Summary

- resolve fff cargo and rustc through the configured rustup toolchain
- install macOS targets into that same toolchain
- add a hermetic regression test for the x86_64 build path

## Root cause

`build-fff.sh` used the bare `cargo` on PATH. On Homebrew Rust installations this selected the native-only toolchain, so the x86_64 slice of universal builds could not find its standard library.

## Validation

- `bash scripts/tests/build-fff/run.sh`
- `xcodegen`

The full Xcode test target is blocked in this worktree because `ThirdParty/ghostty` is uninitialized and `GhosttyKit.xcframework` is unavailable.

Closes #804

### DIFF
--- a/scripts/build-fff.sh
+++ b/scripts/build-fff.sh
@@ -99,6 +99,7 @@ esac
 
 mkdir -p "${lib_dir}" "${include_root}"
 
+"${rustup_bin}" toolchain install "${rust_toolchain}" --profile minimal
 installed_targets="$("${rustup_bin}" target list --installed --toolchain "${rust_toolchain}")"
 if ! printf '%s\n' "${installed_targets}" | grep -qx "${cargo_target}"; then
     "${rustup_bin}" target add --toolchain "${rust_toolchain}" "${cargo_target}"
--- a/scripts/build-fff.sh
+++ b/scripts/build-fff.sh
@@ -11,6 +11,8 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 script_path="${script_dir}/$(basename "${BASH_SOURCE[0]}")"
 srcroot="${SRCROOT:-$(cd "${script_dir}/.." && pwd)}"
 fff_src="${srcroot}/ThirdParty/fff"
+rustup_bin="${ALAS_RUSTUP_BIN:-rustup}"
+rust_toolchain="${ALAS_RUST_TOOLCHAIN:-stable}"
 
 if [ -z "${ALAS_FFF_TARGET_ARCH:-}" ] && [ "${CURRENT_ARCH:-}" = "undefined_arch" ]; then
     target_arch="universal"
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #811, opened 2026-07-16
fix(worktree): keep new worktrees creating until ready

mrmans0n:
## Summary

This fixes the transient massive changes snapshot that can appear while creating a new linked worktree in large repositories such as `android_stb`.

Root cause: topology refresh can see the new worktree in `git worktree list` before checkout has fully materialized files. `ProjectsManager.refreshWorktrees` treated that Git visibility as creation completion and cleared `.creating`, which allowed the right pane to run status/diff against a partial checkout.

## Changes

- Preserve `.creating` during topology refresh reconciliation, while still replacing optimistic row data with the canonical live Git row.
- Clear `.creating` from the owning `AppState.createWorktree` task after checkout, optional startup script, and final refresh complete.
- Add regression coverage for live creating reconciliation preserving operation state while adopting canonical metadata.
- Document the race evidence, lifecycle design, and implementation plan.

## Verification

- `xcodegen` passed locally.
- `git diff --check` passed locally.
- Source review completed with no remaining issues.
- Local `xcodebuild` build/test attempts are currently blocked before compilation by Xcode reporting `There is no XCFramework found at .build/ghostty/GhosttyKit.xcframework`, even after running `scripts/build-ghostty.sh` to populate the ignored build artifact. CI should provide the authoritative full Xcode build/test result for this branch.

### DIFF
--- a/Alas/Sources/App/AppState.swift
+++ b/Alas/Sources/App/AppState.swift
@@ -991,6 +991,7 @@ final class AppState {
                     )
                     let gcDropped = try await projectsManager.refreshWorktrees(projectId: project.id)
                     guard projects.contains(where: { $0.id == projectId }) else { return }
+                    projectsManager.setOperationState(id: optimistic.id, state: nil)
                     if wasHidden || gcDropped {
                         saveProjects()
                     }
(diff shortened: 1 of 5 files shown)


mrmans0n:
Codex review: thumbs up.

I reviewed the lifecycle change and regression coverage. The implementation keeps `.creating` owned by the create workflow, preserves it during topology refresh while reconciling canonical live row metadata, and clears it only after the final successful refresh. CI `build-test` is green for head commit `609c2e088a2ce90d84118772f92b9519be5185ad`.

PR #814, opened 2026-07-16
fix(diff): keep expand pill aligned with text

mrmans0n:
## Summary

Fixes a diff expand-control layout bug where the `Expand 10 unchanged lines` label could appear outside its pill in tall synchronized split rows.

The root cause was that the pill background was vertically centered in the full row rect, while the label text is drawn in the first text line fragment. Split panes can inflate one side's row height to match the other side, which left the pill detached from the text.

## Changes

- Anchor expand pill geometry to the first line fragment used for text drawing, with a fallback to the row rect.
- Reuse that geometry in the AppKit drawing path so the chevron, label, and pill stay together.
- Add a regression test for tall synchronized rows where the row center differs from the text fragment center.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/DiffPaneTextDocumentBuilderTests/expandPillStaysWithFirstTextFragmentInTallRows -quiet test`

Full local suite was also run before rebasing; it failed outside this change on SSH integration tests requiring `ALAS_SSH_INTEGRATION=1` and `RightPaneStateBaseBranchTests/storeRefreshesWhenConfigMatchesClearedOverride()`.

### DIFF
--- a/Alas/Sources/Center/Diff/DiffPaneTextDocumentView.swift
+++ b/Alas/Sources/Center/Diff/DiffPaneTextDocumentView.swift
@@ -1043,6 +1043,40 @@ final class DiffPaneCodeTextView: NSTextView {
         return 0.18
     }
 
+    nonisolated static func expandPillRect(
+        textRect: NSRect,
+        firstLineRect: NSRect,
+        rowRect: NSRect,
+        chevronSize: NSSize,
+        chevronGap: CGFloat = 5,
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #815, opened 2026-07-16
fix(acp): avoid menu tracking list invalidation

mrmans0n:
## Summary

- move modern row-frame bookkeeping out of SwiftUI-observed value state
- ignore stable geometry reports and prune stale rows once
- add regression coverage for stable and stale row-frame cache updates

## Root cause

Each modern row geometry callback assigned the top-level `modernRowFrames` `@State` dictionary, including when the effective frame set was unchanged. AppKit's nested `NSMenu` tracking loop could repeatedly service those callbacks and force a full `ACPMessageList` relayout.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet test`


mrmans0n:
@codex review

PR #820, opened 2026-07-17
fix(diff): render diff expansion actions inline

mrmans0n:
## Summary
Multiple context expansion actions can appear on the same diff row. This PR renders them as separate inline targets so the chunk expansion and full expansion controls sit side by side instead of stacking on consecutive rows.

## Changes
- Tracks each context expansion action with its own text range, boundary, and action mode.
- Draws hover, pressed state, chevron, and pill backgrounds per inline expansion action.
- Dispatches clicks based on the exact inline target, so `Expand all context` invokes full expansion without requiring Option-click.
- Covers the side-by-side builder output and second inline action hit testing in `DiffPaneViewTests`.

## Validation
- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/DiffPaneViewTests test`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -skip-testing:AlasTests/SSHIntegrationTests test`
- `git diff --check`

Note: the unskipped full local suite only failed on `SSHIntegrationTests`, which are explicitly gated by `ALAS_SSH_INTEGRATION=1`.

### DIFF
--- a/Alas/Sources/Center/Diff/DiffContextExpansion.swift
+++ b/Alas/Sources/Center/Diff/DiffContextExpansion.swift
@@ -368,20 +368,7 @@ enum DiffContextExpandedDisplayBuilder {
                 key: key,
                 edge: edge
             )
-            guard edge == .top else {
-                return [expandable]
-            }
-            return [
-                expandable,
-                expandableRow(
-                    group: attachedGroup,
-                    boundary: attachedBoundary,
-                    remaining: 0,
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #821, opened 2026-07-17
fix(diff): Improve diff comment send button

mrmans0n:
## Summary
- Restyle the diff draft-comment Send action as a compact primary button.
- Use a paperplane icon and accent treatment for both single-target and multi-target send controls.
- Preserve existing accessibility identifiers and press markers for automation.

## Validation
- xcodegen
- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test (fails only on SSHIntegrationTests requiring ALAS_SSH_INTEGRATION=1)


mrmans0n:
@codex review

PR #822, opened 2026-07-17
feat(review): support keyboard range selection

mrmans0n:
## Summary
- add Shift+Up/Shift+Down range extension in the review target palette commit list
- reuse the existing commit range target flow when activating the selected range
- fix target-list loading selection so loaded commits snap to the first selectable row

## Verification
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' -only-testing:AlasTests/ReviewTargetPaletteModelTests test`
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' -quiet build`


mrmans0n:
@codex review

PR #823, opened 2026-07-17
fix(acp): transcript live-lock (beachball) in chat view

mrmans0n:
## Summary

Diagnosed and fixed a main-thread live-lock (sustained 100% CPU, beachball) in `ACPMessageList`, the ACP chat transcript view. Captured live via `sample` on a hung instance: the main thread was stuck entirely inside `GraphHost.flushTransactions()`, which never drains because scroll/geometry callbacks write `@State`, and any `@State` write invalidates the view even when body never reads it — each transaction's side effects were enqueuing the next transaction. The driving agent process was idle throughout, confirming this was self-sustaining rather than load-driven.

- Moved callback-only bookkeeping (restore/anchor state) out of observed `@State` into non-observed reference boxes, breaking the core feedback loop
- Memoized the visible-row lookup per render-window generation (was O(rows) per row callback, O(rows²) per layout pass)
- Skipped anchor-tracking work entirely while following the tail (streaming), where it was always discarded
- Stopped misclassifying stale `NSApp.currentEvent` as live user scroll gestures
- Made programmatic tail-scrolls and remembered-anchor restores idempotent/non-re-entrant
- Throttled `streamingTick` publishes to ~30Hz with a trailing-edge guarantee
- Added an `Equatable` gate on transcript rows so unchanged rows skip SwiftUI's subtree re-diff (the dominant cost in the live sample), with a stale-closure audit to confirm no excluded dependency can go stale
- Removed a per-row O(rows) stale-key scan in the row-frame cache that was still running unconditionally on every callback

## Test plan

- [x] Full test suite passes locally (5693 tests; only the 8 pre-existing `ALAS_SSH_INTEGRATION`-gated SSH integration tests fail, unrelated to this change)
- [x] Each commit built and tested independently before merge
- [ ] Manual: long transcript + active streaming — tail-follow stays pinned, pause/resume on scroll works, head/bottom pagination work, remembered anchor survives tab switch
- [ ] Manual: idle CPU stays low after streaming ends (previously pegged at 100% indefinitely)


mrmans0n:
@codex review

PR #825, opened 2026-07-17
feat(acp): make injected MCP tools discoverable via a first-prompt preamble

mrmans0n:
## Problem

Agents run inside Alas (Claude, Codex, Cursor) reported having no `alas` MCP tools and rarely used them, even though the MCP badge showed the server attached.

Root cause (verified with a minimal ACP client + a logging shim around `alas mcp`): the plumbing works end-to-end — every agent receives the `mcpServers` payload, spawns `alas mcp`, completes the handshake, and fetches all 15 tools via `tools/list`; explicit tool calls succeed. But **modern harnesses defer MCP tools behind tool search** (Codex core 0.144.x exposes only `tool_search`; Claude Code defers when many tools are attached), so the model never sees the tools and doesn't reach for them. The server-level `instructions` string isn't surfaced to the model by Codex at all. Any user-configured MCP server injected by Alas suffers identically.

## Fix

Inject a short, one-time, **wire-only** context preamble into the first prompt of each freshly created ACP session, naming the attached MCP servers and telling the agent to use its tool-search mechanism to load them. It never enters the recorded transcript; the MCP badge popover gains an informational "preamble pending/sent" line.

- **`ACPMCPPromptPreamble`** — pure builder; tool-name list mirrors `mcp.rs` `tool_definitions` (drift-guarded by test). Root vs delegated variants; names user-configured servers too.
- **Schema v13** — persisted `mcp_preamble_pending` / `mcp_preamble_sent` (additive, idempotent, preserved across upserts) so the preamble survives an app restart before the first prompt.
- **Attach** — computed and stored only on the fresh `session/new` path (never on load/resume/fork), writer-lease-guarded; restored on hydrate.
- **Runner** — prepends the preamble to the first wire prompt, clears + persists on success, keeps it pending on failure.
- **Badge popover** — shows preamble pending/sent.

Local-only by construction: for remote SSH sessions the built-in stdio server is skipped, so the preamble omits the built-in line but still names surviving http/sse user servers.

## Verification

- Build clean; 207/207 across the 10 touched focused suites, plus the added remote-composition test.
- **End-to-end A/B against `codex-acp`**, identical bare prompt ("What git worktrees exist in this project?"):
  - **With** the preamble → codex plans a tool search, finds `mcp.alas.worktree_list`, and **invokes it**.
  - **Without** → codex uses only shell tools; the `alas` MCP server is never touched.

## Notes

- No changes to the Rust `alas mcp` server or its tool descriptions.
- The MCP badge still reflects the attach plan (not observed connections) by design; a popover line now surfaces preamble delivery state.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -978,6 +978,7 @@
 		C245DDADA1D81810E9302134 /* VerdictSheet.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2942AD875C0E7AB71E8E3CAD /* VerdictSheet.swift */; };
 		C2CE2E3DE88344FDDE67D66A /* LSPTransportTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 08ACB3B79B74153668DA29D8 /* LSPTransportTests.swift */; };
 		C2EDB6A5E6149758C4A0051E /* ConflictMarkerParser.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2392A88D394E21DAB8C6EF33 /* ConflictMarkerParser.swift */; };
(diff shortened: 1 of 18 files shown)


mrmans0n:
@codex review

PR #827, opened 2026-07-17
fix(diff): break split-pane row-height sync live-lock (beachball)

mrmans0n:
## Summary
- Fixes a live-lock (beachball) in split-layout diff review: `synchronizeRowHeights` stripped a row's custom line-height padding whenever any *other* row's target height changed, even when that row's own padding was still needed.
- Removing the padding shrank the row, which the next layout pass detected as a change, reapplying the padding and invalidating layout again — an unbounded apply/strip/reapply cycle that pegged the main thread at 100% CPU for any split diff with a line-wrap mismatch between old/new sides (a very common case, not an edge case).
- Root-caused from a live `sample` trace of the actually-hung process (main thread stuck continuously in `GraphHost.flushTransactions()` across multiple samples over 30 minutes with no user input).

## Test plan
- [x] Updated `synchronizedRowHeightsResetPreviouslyPaddedRowsForRemeasurement` (renamed `synchronizedRowHeightsKeepsStablePaddingWhileOtherRowsChange`) to assert the correct, stable behavior instead of the oscillation it previously encoded as intended.
- [x] Added `synchronizeRowHeightsIsNoOpOnceStable`, a direct regression guard asserting a repeat call with unchanged heights does not re-trigger `needsLayout`.
- [x] `xcodebuild test` — full suite passes (only pre-existing, opt-in `SSHIntegrationTests` gated behind `ALAS_SSH_INTEGRATION=1` fail, unrelated).

### DIFF
--- a/Alas/Sources/Center/Diff/DiffPaneTextDocumentView.swift
+++ b/Alas/Sources/Center/Diff/DiffPaneTextDocumentView.swift
@@ -794,9 +794,10 @@ final class DiffPaneTextScrollView: NSScrollView {
             // otherwise leave a stale per-line height in place and make the
             // row too short to match its counterpart.
             var candidateLineHeight: CGFloat?
+            var lineCount = 1
             if let targetHeight {
                 let lineCountHeight = previousLineHeight ?? lineHeight()
-                let lineCount = max(1, Int(round(currentRows[index].height / max(lineCountHeight, 1))))
+                lineCount = max(1, Int(round(currentRows[index].height / max(lineCountHeight, 1))))
                 let computed = targetHeight / CGFloat(lineCount)
                 if computed > lineHeight() + 0.5 {
                     candidateLineHeight = computed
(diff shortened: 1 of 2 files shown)


mrmans0n:
Good catch — pushed a follow-up that re-derives the per-line height from the row's current geometry every time instead of gating on "did the target change", so a resize-driven wrap-count change gets picked up even when the target height coincidentally stays the same. Added a regression test for that exact scenario.

@codex review

mrmans0n:
Both good catches too — fixed:
- Tolerance now compares total row-height delta (per-line delta × line count), not per-line delta alone, so a small-per-line-but-large-in-aggregate change across many wraps no longer gets skipped.
- `lineTones`/`theme` are only forwarded to the text view when they actually changed, so a fully-stable sync no longer forces a row-geometry invalidation on every parent layout pass.

@codex review

mrmans0n:
Good one — the lineTones-unchanged guard did have that side effect. Now explicitly invalidating row geometry whenever a paragraph style actually changed, verified with a test that reads `diffRowRects()` immediately after `synchronizeRowHeights` (no intervening layout pass) to catch the exact staleness window.

@codex review

PR #826, opened 2026-07-17
feat(acp): Restart local ACP sessions through broker

mrmans0n:
## Summary
- add helper-backed ACP broker state and process supervision for recoverable local sessions
- route local ACPSessionManager attaches through ACPBrokerClient and persist broker cursors
- replay adopted broker initialize/session snapshots after relaunch and harden helper environment handling

## Testing
- rtk git diff --check
- rtk cargo test --manifest-path AlasHelper/Cargo.toml
- rtk xcodegen
- rtk env ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/ACPSessionStoreSchemaTests -only-testing:AlasTests/ACPSessionStoreCRUDTests -only-testing:AlasTests/ACPBrokerClientTests -only-testing:AlasTests/ACPSessionManagerAttachRestoreTests -quiet test
- rtk env ALAS_FFF_TARGET_ARCH=arm64 xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build
- Full xcodebuild test before rebase ran 5689 tests; only SSHIntegrationTests failed because ALAS_SSH_INTEGRATION=1 is required.

PR #829, opened 2026-07-17
fix(acp): preserve integer params through broker JSON round-trip

mrmans0n:
## Problem

Since v0.11.5, attaching to ACP chat sessions fails with `ACP session attach failed: ACP error -32602: Invalid params`, across **all providers** (codex, claude, opencode) and for **most chats**. It did not happen in v0.11.4.

## Root cause

The local broker (introduced in #826) routes every request through `ACPBrokerClient`, which re-encodes outgoing params via `JSONSerialization` in `ACPBrokerJSONValue.init(jsonObject:)`. That `switch` matched `Bool` before `Int`:

```swift
case let value as Bool:
    self = .bool(value)
case let value as Int:
    self = .number(Double(value))
```

`JSONSerialization` boxes every JSON number as an `NSNumber`, and Foundation's `NSNumber → Bool` bridge succeeds for any value equal to `0` or `1`. So the `initialize` request's `protocolVersion: 1` was rewritten to `true`. Every conformant ACP agent validates `protocolVersion` as a number, so it rejected the request with `-32602` — before `session/new` was ever reached. Because the broker is the mandatory transport for all locally-spawned agents since #826, every fresh `initialize` was affected regardless of provider.

`mcpServers` (including the built-in `alas` server) is unaffected — it's all strings and round-trips losslessly.

## Fix

Collapse the numeric cases into a single `NSNumber` case and use `CFBoolean`'s distinct `CFTypeID` to tell genuine JSON booleans apart from numbers before mapping. Booleans stay boolean; `0`/`1`/`2`… stay numbers.

## Tests

- `encodableIntegerParamsSurviveAsNumbers` — regression guard: `protocolVersion: 1` encodes to `.number(1)`, not `.bool(true)`.
- `encodableBooleansStayBooleans` — ensures the fix doesn't over-correct real booleans.

`AlasTests/ACPBrokerClientTests` passes 25/25 locally (arm64).

### DIFF
--- a/Alas/Sources/ACP/Broker/ACPBrokerClient.swift
+++ b/Alas/Sources/ACP/Broker/ACPBrokerClient.swift
@@ -745,7 +745,7 @@ private extension ACPBrokerAdapterRequestID {
     }
 }
 
-private extension ACPBrokerJSONValue {
+extension ACPBrokerJSONValue {
     init(encodable value: Encodable?) throws {
         guard let value else {
             self = .null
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #830, opened 2026-07-18
fix(integrations): code-host authentication for SSH workspaces 

mrmans0n:
## Summary

- route GitHub and GitLab CLI commands through the registered SSH host for remote workspaces
- make provider availability checks workspace-aware
- preserve locale, git lock, argument quoting, and stdin behavior across remote execution
- add regression coverage for SSH command routing

## Root cause

The changes pane detected the remote repository correctly, but launched `gh` and `glab` locally with the remote filesystem path as `cwd`. That process launch failed before the authentication probe ran, and the failure was surfaced as **Auth needed**.

## Impact

GitHub and GitLab status, review, and action commands now execute on the same remote host as the SSH workspace. Local workspaces retain the existing local execution path.

## Validation

- `xcodegen`
- focused provider suites: 167 tests passed
- non-opt-in full suite: 5,785 tests passed, 4 skipped

The unfiltered suite additionally discovers eight localhost SSH integration tests that are intentionally gated on `ALAS_SSH_INTEGRATION=1`; Xcode does not forward that shell variable into this scheme's test host.

### DIFF
--- a/Alas/Sources/Integrations/CodeHost/CodeHostProvider.swift
+++ b/Alas/Sources/Integrations/CodeHost/CodeHostProvider.swift
@@ -40,13 +40,44 @@ extension CodeHostCommandRunning {
 
 struct ProcessCodeHostCommandRunner: CodeHostCommandRunning {
     func run(_ executable: String, args: [String], cwd: URL?, stdin: String?) async throws -> ProcessResult {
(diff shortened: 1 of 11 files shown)


mrmans0n:
@codex review

PR #831, opened 2026-07-18
fix(acp): preserve brokers across helper restarts

mrmans0n:
## Summary

- add an explicit process-only termination scope to the JSON-RPC stdio transport
- use that scope for the local ACP broker-control helper
- preserve the existing process-tree cleanup behavior for all other transports
- add regression coverage for the local broker transport configuration

## Root cause

The local helper launches each durable ACP broker supervisor as a detached descendant. The generic stdio transport still tracked that supervisor and killed it individually when the helper channel exited or restarted. A subsequent helper process found the stale broker socket and failed with `Remote helper error -32072: broker connect failed: Connection refused (os error 61)`.

## Impact

Local ACP chats no longer lose their broker supervisor when the control helper restarts, so in-flight and persisted sessions remain reachable across helper channel churn.

## Validation

- `xcodegen`
- focused `ProcessTransportTerminationTests` pass
- AlasHelper Rust suite: 76 passed
- full macOS test run: 5,786 passed; only the 8 opt-in `SSHIntegrationTests` failed because `ALAS_SSH_INTEGRATION=1` was not set

### DIFF
--- a/Alas/Sources/ACP/Broker/LocalACPBrokerService.swift
+++ b/Alas/Sources/ACP/Broker/LocalACPBrokerService.swift
@@ -37,7 +37,8 @@ actor LocalACPBrokerService {
                     executable: binary,
                     arguments: ["serve"],
                     environment: nil,
-                    framing: .newline
+                    framing: .newline,
+                    terminationScope: .processOnly
                 )
             }
         )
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #832, opened 2026-07-18
fix(remote-web): show loading screen instead of disconnected flash on initial load

mrmans0n:
## Problem

On the remote web, loading an already-paired site flashed the alarming **"Can't reach Alas / Try again"** disconnected screen for a few seconds before content appeared.

Root cause: `socket.onclose` showed that gate unconditionally. Per the WebSocket spec, `onclose` fires even when the initial handshake never completed — so on a fresh load, a slightly slow `/ws` connection (mobile radio waking, LAN/tailnet latency, server not yet accepting) closed the first socket before `onopen`, painting the disconnected screen immediately. The reconnect backoff then held it there until a retry succeeded.

## Fix

Distinguish **loading** (never connected yet this page load) from **disconnected** (was connected, then dropped), via an `everConnected` flag, and escalate to the alarming gate only after a ~5s grace period (a *total* continuous-not-connected budget, not per-retry).

- **Normal load** → brief neutral "Connecting…" spinner → content. No alarming flash.
- **Mac genuinely unreachable at load** → spinner for ~5s → "Can't reach Alas / Try again", and it stays put.
- **Mid-session blip** → transcript stays visible, only a "Reconnecting…" status chip; the full gate appears only after ~5s of sustained failure.
- **"Try again"** → back to the spinner with a fresh grace budget.

Implemented with a single escalation timer (`GRACE_MS = 5000`), an `escalated` one-way latch (reset on a successful connect or manual retry), a `showConnectingGate()` variant reusing the existing gate overlay, a pure `closeState()` helper, and a `#gate.connecting` spinner in CSS. Service-worker `CACHE_NAME` and pinned asset versions bumped so installed PWAs pick up the new `app.js` / `style.css`.

## Testing

The remote-web client is vanilla JS with no JS test harness in the repo (native tests use Swift Testing), so verification was `node --check` (syntax) plus code-level review of the connection state machine. The following runtime scenarios should be walked in the app before merge:

1. Normal load (Mac up) → brief "Connecting…", no alarming flash.
2. Mac off at load → "Connecting…" ~5s → "Can't reach Alas / Try again" (stays stable).
3. Mid-session Wi-Fi blip → transcript stays, "Reconnecting…" chip, recovers with no overlay.
4. Mid-session sustained drop → chip ~5s → full gate over content.
5. "Try again" → returns to "Connecting…", restarts the budget.

## Out of scope

`pairWithCode()`'s `/pair` network-failure still shows the alarming gate immediately — that's the QR-pairing flow, not the already-paired load path this PR targets.

### DIFF
--- a/Alas/Resources/RemoteWeb/app.js
+++ b/Alas/Resources/RemoteWeb/app.js
@@ -18,6 +18,9 @@ let pairingPromise = null;
 let pairingController = null;
 const initialReconnectDelay = 1500;
 const maxReconnectDelay = 30000;
+let everConnected = false;      // has any onopen fired this page load? separates "loading" from "disconnected"
+let escalationTimer = null;     // fires after a continuous not-connected grace window, then shows the alarming gate
+const GRACE_MS = 5000;          // total not-connected budget before escalating "Connecting…" → "Can't reach Alas"
 let dismissedQuestion = null;   // {sessionId, requestId} the user closed; suppress re-shows of that exact prompt (ids aren't unique across sessions)
 let lastSentText = null;        // text of the most recent sendPrompt, kept so a server promptRejected can restore it instead of losing the message
 let lastSentAttachments = [];   // images of the most recent sendPrompt, restored alongside the text on promptRejected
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #838, opened 2026-07-18
fix(acp): throttle streaming text row publishes to display rate

### LINKED ISSUE #835: Streaming markdown tail is re-parsed per chunk and never promotes inside fenced code blocks
## Problem

While an agent streams a reply, the un-promoted markdown tail is fully re-parsed on every text chunk, and the tail can grow without bound:

- `StreamingText.value` is `@Published` and appended per streaming chunk with no throttle (`Alas/Sources/ACP/Session/ACPStreamingText.swift`). The #823 fix throttled `streamingTick` (whole-list invalidation) but the per-row buffer publish still fires per chunk.
- On each publish, the agent message row re-runs `ACPMarkdownText.parse(cache.tailUnparsed)` over the entire un-promoted tail.
- `ACPMarkdownBlockCache.promoteIfPossible()` (`Alas/Sources/ACP/UI/ACPMarkdownBlockCache.swift:112`) only promotes the stable prefix at a blank line **outside** fenced code blocks. While streaming a long fenced block (or any long blank-line-free stretch), the tail never promotes → O(tail) parse per chunk → O(tail²) total.
- The rendered blocks use a non-lazy `VStack` + `ForEach(...enumerated(), id: \.offset)`, so every chunk also rebuilds view values for **all** stable blocks of the message, O(blocks) per chunk.

## Impact

For a multi-hundred-KB streamed reply containing large code blocks, this is the dominant per-chunk main-thread cost in the transcript — degrades toward a beachball as the reply grows, same family as #823 via throughput.

## Suggested direction

- Throttle/coalesce `StreamingText` publishes to display rate, mirroring the `streamingTick` throttle from #823.
- Add an in-fence promotion point (e.g. promote completed lines within a fence as a provisional code block, or track fence-open state so the stable prefix can advance line-by-line inside fences).
- Make the block list lazy or identity-keyed so stable blocks skip re-evaluation per chunk.

mrmans0n:
## What

Throttle the per-row `StreamingText` publish to display rate so a streaming
agent reply no longer re-parses its markdown tail on every chunk.

Closes #835.

## Why

`AgentMessageRow` (and `ACPThoughtView`) observe `StreamingText` via
`@ObservedObject`, and `StreamingText.value` was `@Published`. Every streaming
chunk therefore published a change and forced a full row re-render, which
re-runs `ACPMarkdownBlockCache.update` and `ACPMarkdownText.parse(tail)` over
the entire un-promoted tail. A fast agent emits chunks far faster than the
display refreshes, so this fired potentially hundreds of times per second — the
dominant per-chunk main-thread cost described in the issue.

The #823 fix throttled `streamingTick` (whole-list invalidation) but not this
per-row buffer publish, which is exactly what the issue calls out.

## How

- `value` is now a plain stored property. It still mutates synchronously on
  every `append`, so any reader (e.g. `ACPSession`, `ACPThoughtView`) always
  sees the latest text immediately.
- The SwiftUI-facing `objectWillChange` publish is throttled to display rate,
  reusing the already-tested pure decision function
  `ACPTranscript.streamingTickAction` from #823. A trailing-edge drain
  guarantees the final chunk of a burst still renders within one throttle
  interval, so no text is ever lost or left unrendered.

This coalesces a burst of chunks into at most ~30 row re-renders/sec instead of
one per chunk, mirroring how #823 was resolved.

## Scope note

The issue lists two further directions — in-fence stable promotion and a
lazy/identity-keyed block list. This PR intentionally lands only the throttle,
which is the dominant, low-risk fix and directly targets "the per-row buffer
publish still fires per chunk":

- **In-fence promotion** cannot preserve the single rendered code block without
  either splitting it into multiple cards (a visual regression) or an
  append-only code renderer (a larger architectural change that per repo
  guidelines needs an explicit plan). Because a growing live code block must be
  re-materialised to display it, promotion yields no per-frame asymptotic win
  once publishes are already throttled.
- The block list already uses positional identity (`id: \.offset`), which is
  stable for the frozen prefix; switching to `LazyVStack` risks scroll
  regressions for a marginal gain after throttling.

Happy to follow up on either in a separate PR if desired.

## Testing

- New `ACPStreamingTextTests`: a synchronous burst of 200 appends coalesces to a
  single publish while `value`/`revision` still reflect every chunk; the
  trailing chunk of a burst is delivered via the drain.
- Existing `ACPStreamingTextTests`, `ACPMarkdownBlockCacheTests`,
  `ACPTranscriptTests`, `ACPTranscriptChangeLogTests`,
  `ACPMessageListPaginationTests` pass locally (arm64).

### DIFF
--- a/Alas/Sources/ACP/Session/ACPStreamingText.swift
+++ b/Alas/Sources/ACP/Session/ACPStreamingText.swift
@@ -11,17 +11,42 @@ import Combine
 /// share the same buffer compare equal even as the text grows, so the
 /// transcript-array diff doesn't fire on streaming. The inner row view
 /// observes the buffer directly via @ObservedObject and re-renders
-/// when `value` changes.
+/// when the buffer publishes.
+///
+/// `value` is NOT `@Published`: the string mutates synchronously on every
+/// `append` (so any reader sees the latest text immediately), but the
+/// SwiftUI-facing `objectWillChange` publish is throttled to display rate
+/// via `ACPTranscript.streamingTickAction`. Without this, a fast agent
+/// response forces one full agent-row re-render — which re-runs
+/// `ACPMarkdownBlockCache.update` + `ACPMarkdownText.parse(tail)` — per
+/// streaming chunk (potentially hundreds/sec). Mirrors the `streamingTick`
+/// throttle from #823, which coalesces whole-list invalidation but does
+/// not cover this per-row buffer publish. A trailing-edge drain guarantees
+/// the final chunk of a burst still renders within one throttle interval.
 @MainActor
 final class StreamingText: ObservableObject {
-    @Published private(set) var value: String
+    private(set) var value: String
     private(set) var utf8Length: Int
     private(set) var revision: UInt64 = 0
     private(set) var lastAppendedSuffix: String?
 
+    /// Wall-clock time (`ProcessInfo.systemUptime`) of the last publish.
+    private var lastPublish: TimeInterval = 0
+    /// Trailing-edge publish scheduled when a chunk arrives inside the
+    /// throttle window; guarantees the last chunk of a burst renders.
+    private var drainTask: Task<Void, Never>?
+
+    #if DEBUG
+    private(set) var publishCountForTests = 0
+    #endif
+
     init(_ initial: String = "") {
         self.value = initial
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #839, opened 2026-07-18
fix(acp): bound live terminal tail rendering

### LINKED ISSUE #834: ACP terminal tail re-parses the entire buffer on every pipe chunk
## Problem

`ACPTerminalTailView` re-renders on every `ACPTerminal.buffer` publish, and each render re-parses the whole buffer from scratch:

- `ACPTerminal.buffer` is a `@Published Data` appended per readability-handler chunk with no coalescing (`Alas/Sources/ACP/Terminal/ACPTerminal.swift`), so a chatty subprocess (build, test run) publishes tens to hundreds of times per second.
- `TerminalLiveBody.parsedAttributed()` (`Alas/Sources/ACP/UI/ACPTerminalTailView.swift:50`) runs a fresh `ANSIStream().feed(terminal.buffer)` over the **entire** buffer (cap 1 MiB), then rebuilds an `AttributedString` with repeated `out += s` (quadratic in run count).
- The result is laid out as a single selection-enabled monospaced `Text` holding up to 1 MiB of content, inside the transcript row.
- `.onChange(of: terminal.buffer)` (line 39) additionally performs an O(buffer) `Data` equality compare and a `scrollTo` per chunk.
- After the cap is hit, `appendChunk`'s `buffer.removeFirst(drop)` is an O(1 MiB) memmove per chunk, on the main actor.

## Impact

Per-chunk cost is O(current buffer) → O(n²) over the life of the process output until the 1 MiB cap, then a constant ~1 MiB parse + layout per chunk indefinitely. With an expanded tool-call terminal row streaming a chatty command, the main thread saturates — same beachball family as #823/#827, via throughput rather than a feedback loop.

## Suggested direction

- Coalesce buffer publishes to display rate (the pattern already used for `streamingTick` in #823).
- Parse incrementally: keep the `ANSIStream` state and parsed runs for the stable prefix, feed only appended bytes.
- Consider rendering only the visible tail (the view is capped at 300 px anyway) instead of the full 1 MiB backlog.

mrmans0n:
Closes #834.

## What changed

- parse ACP terminal output incrementally instead of reparsing the retained 1 MiB buffer on every pipe chunk
- retain a bounded 64 KiB parsed display tail and publish UI revisions at up to 30 Hz with trailing delivery
- build the terminal `AttributedString` from one joined string and apply attributes by range
- make rolling protocol-buffer compaction amortized instead of shifting 1 MiB per chunk after rollover
- add coverage for split ANSI and UTF-8 sequences, cross-chunk carriage-return redraws, tail bounds, and display-update coalescing

## Impact

Chatty subprocess output no longer causes per-chunk O(buffer) parsing, equality checks, scrolling, and layout. The full 1 MiB protocol tail remains available to `terminal/output`, while the 300-point transcript row renders a bounded display tail.

## Validation

- `xcodegen`
- focused `ANSIStreamTests` and `ACPTerminalTests`
- full test suite: 5,792 passed and 4 skipped; the remaining 8 failures are the opt-in `SSHIntegrationTests` requiring `ALAS_SSH_INTEGRATION=1`

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSession.swift
+++ b/Alas/Sources/ACP/Session/ACPSession.swift
@@ -1163,7 +1163,7 @@ final class ACPSession: ObservableObject, Identifiable {
         if replace { return true }
         if replayCreatedMetadataTerminalIds.contains(terminalId) { return true }
         if let terminal = terminalHost.terminal(id: terminalId),
-           !terminal.buffer.isEmpty || terminal.exitStatus != nil {
+           terminal.retainedByteCount > 0 || terminal.exitStatus != nil {
             return false
         }
         replayCreatedMetadataTerminalIds.insert(terminalId)
--- a/Alas/Sources/ACP/Terminal/ACPTerminal.swift
+++ b/Alas/Sources/ACP/Terminal/ACPTerminal.swift
@@ -4,14 +4,23 @@ import Combine
 
 /// Wraps one agent-spawned subprocess. Owns the merged stdout+stderr
 /// rolling buffer (capped at 1 MiB), captures the exit status, and
-/// publishes a "buffer changed" signal so UI subscribers can re-render
-/// the live tail without polling.
+/// publishes a display-rate-limited terminal tail for the transcript UI.
 @MainActor
 final class ACPTerminal: ObservableObject {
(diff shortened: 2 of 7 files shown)


mrmans0n:
@codex review

PR #840, opened 2026-07-18
fix(acp): incremental ACP transcript plan caching

### LINKED ISSUE #837: refreshPlanCaches scans the full transcript on every messages mutation
## Problem

`ACPTranscript.refreshPlanCaches()` runs inside `messages.didSet` (`Alas/Sources/ACP/Session/ACPTranscript.swift:12-16`), i.e. on **every** array mutation: appended rows, every `toolCallUpdate` reassignment (`messages[i] = .toolCall(tc)`), window trims, hydration steps.

The `newLatestPlan` loop (`ACPTranscript.swift:136-142`) breaks only when it finds a `.plan` message. In sessions whose agent never emits plans — common — it walks the **entire** `messages.reversed()` on every single mutation. The `newCurrentPlan` loop similarly walks back to the last `.user` message, which is O(current turn length) for long tool-heavy turns.

## Impact

O(n) enum scans per mutation × per-update mutation rate → O(n²) growth as the transcript gets long. A 10k-row transcript in a busy tool-streaming turn burns hundreds of thousands of enum compares per second on the main actor, for a cache nothing may even be observing. Contributes to main-thread saturation in the same sessions affected by #823.

Compounding factor: window-trim loops (`shiftVisibleHeadAfterPrepending` / `setVisibleWindow`, `ACPTranscript.swift:305-339`) write `messages[index]` per trimmed row, so a single hydration step fires this full scan once per trimmed row.

## Suggested direction

- Track the index of the latest `.plan` (and latest `.user`) incrementally as messages are appended/updated, instead of rescanning; or
- Short-circuit when the session has never contained a `.plan` message (a single boolean/count kept in sync makes the no-plan case O(1)); and
- Batch the trim loops into a single array mutation so `didSet` fires once.

mrmans0n:
## Summary

- track the latest plan and user message indices incrementally for production transcript mutations
- use the cached current-plan index when applying plan updates
- batch off-window tool-call truncation into a single transcript publication
- add regressions covering cache rebuild avoidance and batched trimming

## Root cause

`messages.didSet` scanned the transcript backward for both plan caches on every append and element replacement. Tool streaming therefore made cache maintenance O(n) per update, and trim loops multiplied that cost by publishing once per row.

Closes #837

## Validation

- `xcodegen`
- focused ACP transcript plan/window/truncation test suites
- full suite: 5,797 passed and 4 skipped; the remaining 8 local failures were opt-in SSH integration tests requiring `ALAS_SSH_INTEGRATION=1`

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSession.swift
+++ b/Alas/Sources/ACP/Session/ACPSession.swift
@@ -295,7 +295,7 @@ final class ACPSession: ObservableObject, Identifiable {
         // of the user message instead of being dropped. The runner persists
         // from the pre-call message count, so the flushed rows are saved too.
         _ = flushPendingReplayCandidates()
-        transcript.messages.append(.user(
+        transcript.appendMessage(.user(
             id: UUID(),
             messageId: nil,
             text: text,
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

PR #846, opened 2026-07-18
fix(acp): cache streamed message indices

### LINKED ISSUE #842: Each streamed ACP chunk triggers two full-transcript firstIndex scans
## Problem

Every streamed update pays two front-to-back scans of `transcript.messages` on the main actor:

1. `ACPSessionRunner.enqueueIncomingUpdate` → `persistedCandidateIndices` (`Alas/Sources/ACP/Session/ACPSessionRunner.swift:439-511`). While the runner holds the write lease (the normal live-prompt case), `capturePersistedBasesForIncomingUpdate` runs immediately per raw update — the 16 ms coalesce only batches the *apply*, not this. It calls `messages.firstIndex(where:)` matching `messageId` (lines ~468, ~484) or `toolCallId` (line ~499). The `toolCallId` lookup bypasses the `toolCallIndices` cache that `updateToolCall` already maintains (`ACPSession.swift:1985-2004`).
2. `ACPSession.messageIndex` (`Alas/Sources/ACP/Session/ACPSession.swift:1541-1552`), called per chunk from `apply` (lines ~366, ~396) for `agentMessageChunk`/`agentThoughtChunk` with a `messageId` (modern adapters always set one) — a second full forward scan.

Streaming targets are almost always at or near the tail, so the forward scan is worst-case O(n) on essentially every chunk.

## Impact

O(n) per chunk × chunk rate → O(n²) as the transcript grows. A 10k-row transcript at ~100 chunks/s is on the order of a million enum/string compares per second on the main thread, before any rendering work. Contributes to the same main-thread saturation family as #823/#834/#837.

## Suggested direction

- Reuse/extend the existing `toolCallIndices` cache for the runner-side `toolCallId` lookup.
- Maintain an analogous `messageId → index` map (already invalidated by the `messagesGeneration` bump), or scan from the tail backwards since targets are nearly always trailing.

Found during the proactive beachball audit following #823/#827; sibling of #837.

mrmans0n:
## Summary

- add incremental `messageId → index` and `toolCallId → index` caches to `ACPTranscript`
- reuse the shared caches from streamed message application and runner-side persisted-base capture
- keep cache updates O(1) for append/replace mutations and rebuild only for whole-array replacement or prepend hydration
- add regression coverage for restore, streaming, replacement, tool-call, and prepend paths

## Root cause

Each modern ACP agent/thought chunk performed a forward scan in both `ACPSessionRunner` and `ACPSession`. Tool-call persistence capture also bypassed the existing session cache. As transcripts grew, the per-chunk O(n) scans saturated the main actor.

## Impact

Streamed message and tool-call lookups are now O(1) in the steady state, removing the transcript-length multiplier from high-frequency updates.

## Validation

- `xcodegen`
- focused `ACPTranscriptMessageIndexTests`
- `ACPSessionTests`
- full suite: 5,802 tests passed; eight opt-in SSH integration tests require `ALAS_SSH_INTEGRATION=1`
- unrelated `ACPTerminalTests/releaseReachesOrphans` timing failure passed on isolated retry

Fixes #842

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -1234,6 +1234,7 @@
 		F18EB035A8972F21D23F3810 /* SearchFooter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 39DFD138FE855D8CFC464847 /* SearchFooter.swift */; };
 		F1AB5B4DF8A59DF85D585058 /* RemoteWeb in Resources */ = {isa = PBXBuildFile; fileRef = B586C7A794B846F3AD9228BD /* RemoteWeb */; };
 		F1C4DD0D3A95ED5301EE2AC9 /* ACPTranscriptMarkdownCacheEvictionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3708552C2BF0E70F10C76B63 /* ACPTranscriptMarkdownCacheEvictionTests.swift */; };
+		F22A776D791ABA7A584B4C18 /* ACPTranscriptMessageIndexTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44E26C51ACE48A35EB4CD463 /* ACPTranscriptMessageIndexTests.swift */; };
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #849, opened 2026-07-19
fix(acp): incremental toolCallUpdate content processing (#843)

### LINKED ISSUE #843: toolCallUpdate re-processes the full cumulative content on every update
## Problem

Adapters that stream tool output emit repeated `toolCallUpdate`s carrying the **full cumulative** content, and Alas re-processes all of it on each one:

- `ACPSession.apply` → `applyToolCallUpdateFields` (`Alas/Sources/ACP/Session/ACPSession.swift:435-443`, `:638-685`) runs `Self.flatten(content)`, `stripWrappingFence`, `replaceContent`, `previewLine`, `extractAssets`, and `metadataString(rawOutput)` — all O(content size) — per update.
- The subsequent `messages[i]` reassignment fires the full `messages.didSet` chain: generation bump, plan-cache scan (#837), remote diff recording when a gateway is attached, and a `@Published` publish that re-evaluates the `ACPMessageList` body. Note `messages` publishes are **not** covered by the `streamingTick` throttle from #823.

## Impact

For a command producing m bytes across k updates, total work is O(m²/k) — each update is progressively more expensive. A long-running chatty command (huge test log, verbose build) makes late updates arbitrarily costly, all on the main actor, concurrent with the terminal-tail rendering cost tracked in #834.

## Suggested direction

- Diff against the previously applied content and process only the appended suffix (preview line, asset extraction, and fence stripping can all work incrementally or on the tail).
- Cache derived values (preview, assets, stripped fence state) keyed on content length/prefix so unchanged prefixes are skipped.
- Consider coalescing toolCallUpdate applies at display rate the same way text chunks are ticked.

Found during the proactive beachball audit following #823/#827.

mrmans0n:
## Problem

Adapters that stream tool output emit repeated `toolCallUpdate`s carrying the **full cumulative** content, and Alas re-processes all of it on each one — `flatten`, `stripWrappingFence`, `previewLine`, `extractAssets`, `extractTerminalIds`, and `replaceContent` each run O(content) per update, and the `messages.didSet` chain fires on every update.

For a command producing `m` bytes across `k` updates, total work is O(m·k): each update is progressively more expensive. A long-running chatty command (huge test log, verbose build) makes late updates arbitrarily costly, all on the main actor.

Closes #843.

## Fix

Cache the last fully-applied flattened raw content (and the stripped-raw body + opening-fence-unterminated flag) on `ACPMessage.ToolCall`. On each `toolCallUpdate`/`toolCall` content payload:

- **Identical content, same `isFinal`:** skip all content reprocessing (no `replaceContent`, no `previewLine`, no fence strip, no asset/terminal scan).
- **Prefix extension (suffix-only fast path):** the new flattened raw has the previous one as a prefix, the previous apply was not `isFinal`, and items are append-only. Process only the appended suffix:
  - The new stripped-raw body is `appliedStrippedRaw + suffix`, with an opening-fence-unterminated boundary adjustment so a fence line that arrived alone (`"```swift"`) then continued (`"\nlet x = 1"`) doesn't double-count the separator newline.
  - `preview` and `contentLanguage` depend only on the first line, so they are computed once and reused.
  - Assets and terminal ids are scanned only over the items beyond `appliedItemCount`; `mergeAssets`/`mergeTerminalIds` dedup.
  - The `isFinal` trailing-fence strip runs once, on the single update that flips `isFinal` to true.
- **Otherwise** (first apply, divergent content, appending to a finalized snapshot, or post-truncation restore): full reprocess — exactly the legacy path — and refresh the cache.

`truncateForOffWindow` invalidates the cache so a subsequent update restores the full body via the full-reprocess path.

The cache fields are in-memory only: they are not encoded, excluded from `Equatable`/`Hashable`, and restored rows start with an empty cache and take the full-reprocess path once.

## Tests

New tests in `ACPSessionTests`:
- streamed suffix chunks accumulate into the full content
- streamed fence strips across suffix chunks (including the unterminated-opening-fence boundary)
- streamed assets declared across suffix chunks are preserved (with dedup)
- streamed terminal ids declared across suffix chunks preserve declaration order
- identical cumulative content does not bump `contentRevision`
- divergent (non-prefix) content reprocesses fully

All existing `ACPSessionTests`, `ACPTranscriptTests`, `ACPSessionRunnerTests`, `ACPSessionStoreTests`, `ACPMessageTests`, and `ACPMessageWireTests` continue to pass.

### DIFF
--- a/Alas/Sources/ACP/Session/ACPMessage.swift
+++ b/Alas/Sources/ACP/Session/ACPMessage.swift
@@ -250,15 +250,25 @@ enum ACPMessage: Equatable {
         /// until then `appliedStrippedRaw` stays empty.
         var appliedOpeningFenceUnterminated: Bool = false
         /// In-memory-only cache of the previous content items array, used
-        /// to verify that the first `appliedItemCount` items are unchanged
+        /// to verify that the first `appliedItemCount` items' STRUCTURED
+        /// content (images, resource-links, terminals, diffs) is unchanged
         /// before taking the suffix-only fast path. `flatten` ignores
-        /// images/resource-links/terminals, so equal flattened text does
-        /// NOT imply the structured items are the same — an in-place
-        /// replacement (e.g. `[text, image(old)]` -> `[text, image(new)]`)
-        /// would otherwise skip `extractAssets`/`extractTerminalIds`. COW
-        /// keeps the storage cost minimal; the prefix comparison is
-        /// O(appliedItemCount) which is bounded by the items count.
+        /// those item kinds, so equal flattened text does NOT imply the
+        /// structured items are the same — an in-place replacement (e.g.
+        /// `[text, image(old)]` -> `[text, image(new)]`) would otherwise
+        /// skip `extractAssets`/`extractTerminalIds`. Only structured
+        /// items are compared; text items are allowed to grow in place
+        /// (the common cumulative shape `[text("a")]` -> `[text("ab")]`)
+        /// so the suffix path still fires. COW keeps storage cheap.
         var appliedItemsSnapshot: [ACPToolCallContent] = []
+        /// In-memory-only flag: whether `stripWrappingFence` actually
+        /// stripped an opening fence line from `appliedRawContent`. The
+        /// trailing-fence strip in `stripTrailingFenceLine` must only
+        /// run when an opening fence was recognized — ordinary tool
+        /// output whose final line happens to be `` ``` `` must NOT be
+        /// dropped. Mirrors `stripWrappingFence`'s `isOpeningFence`
+        /// guard.
+        var appliedOpeningFenceStripped: Bool = false
 
         /// Replace `content` with its first `truncatedTailBytes` characters
         /// and mark the message as truncated. No-op if already truncated.
@@ -278,6 +288,7 @@ enum ACPMessage: Equatable {
             appliedStrippedRaw = ""
             appliedOpeningFenceUnterminated = false
             appliedItemsSnapshot = []
+            appliedOpeningFenceStripped = false
         }
 
         init(toolCallId: String, title: String, kind: String? = nil,
--- a/Alas/Sources/ACP/Session/ACPMessage.swift
+++ b/Alas/Sources/ACP/Session/ACPMessage.swift
@@ -209,6 +209,47 @@ enum ACPMessage: Equatable {
         /// user expands.
         static let truncatedTailBytes: Int = 4096
 
+        /// In-memory-only cache of the last fully-applied flattened raw
+        /// content (before fence stripping), used by
+        /// `applyToolCallUpdateFields`/`applyToolCallPayloadFields` to
+        /// detect prefix-extension updates and process only the appended
+        /// suffix. Adapters that stream tool output re-emit the full
+        /// cumulative content on every `tool_call_update`, so without
+        /// this cache each update re-runs `stripWrappingFence`,
+        /// `previewLine`, `extractAssets`, and `extractTerminalIds` over
+        /// the whole growing body — O(m·k) total work for m bytes across k
+        /// updates. Never encoded: restored rows start with a nil cache
+        /// and take the full-reprocess path once, which is fine.
+        var appliedRawContent: String? = nil
+        /// In-memory-only count of `ACPToolCallContent` items already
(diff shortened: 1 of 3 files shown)


mrmans0n:
## Addressing Codex review comments

All P2 comments from the previous rounds have been addressed in the latest commits. The comments re-posted on commit `2d55f42c` reference line numbers that point to already-fixed code. Here is a mapping:

### 1. "Scan structured items before skipping identical text" (line 683)
**Fixed.** The `return` at line 683 is guarded by `items.count == tc.appliedItemCount && prefixItemsUnchanged` (lines 664-665). `prefixItemsUnchanged` calls `structuredContentEqual` which compares images, resource-links, terminals, and diffs exactly (text items are treated as equal since the flattened-raw prefix check guards text). An added screenshot (count grows) or in-place replacement (count same, item differs) falls to full reprocess.

### 2. "Scan terminals inserted before the suffix slice" (line 773)
**Fixed.** The suffix path at line 694-698 requires `prefixItemsUnchanged`. A terminal inserted before a text item changes a structured item, so `prefixItemsUnchanged` returns false and the suffix path falls to full reprocess.

### 3. "Guard trailing-fence stripping by opening fence" (line 742)
**Fixed.** Line 742 now reads `if isFinal && tc.appliedOpeningFenceStripped`. `appliedOpeningFenceStripped` is set in `reprocessToolCallContentFull` by checking `isOpeningFence(firstLine)`. Ordinary tool output (no opening fence) does not strip a trailing ` ``` ` line.

### 4. "Let growing text items take the suffix path" (line 698)
**Fixed.** `prefixItemsUnchanged` uses `structuredContentEqual` which returns `true` for `(.content(.text), .content(.text))` without comparing the text. The common shape `[text("a")]` -> `[text("ab")]` takes the suffix path.

### 5. "Recompute assets when rawOutput changes" (line 683)
**Fixed.** The identical-text skip at line 666 requires `!rawOutputChanged && rawOutputAssets.isEmpty`. If `update.rawOutput` is non-nil, `rawOutputChanged` is true and we fall to full reprocess, which rebuilds `tc.assets` from content + current rawOutput.

### 6. "Avoid flattening the cumulative body before the fast path" (line 663)
**Not addressed — by design.** `flatten(items)` is O(content) but it is the cheap part (string concatenation). The expensive operations — `stripWrappingFence` (split + join), `previewLine` (split), `extractAssets` (item scan), `extractTerminalIds` (item scan), `extractStoredRawOutputAssets` (JSON parse) — are what the fast paths skip. Avoiding `flatten` would require caching the items array and comparing it element-wise, which is itself O(items count) and adds complexity for marginal gain. The net improvement is O(m·k) -> O(m) for the expensive scans.

### 7. "Recompute first-line fields until their source line is stable" (line 762)
**Fixed.** `firstLineIsComplete` now returns `false` when the first line is empty or absent (no newline), and `true` only when a non-empty first line is followed by a newline. The preview (first non-empty line) and fence language are recomputed until the first non-empty line is complete.

### 8. "Rebuild terminal ids from the current snapshot" / "Rebuild terminal ids for metadata changes" (line 774, 683)
**Fixed.** The identical-text skip at line 666 requires `!metadataChanged`. If `update.metadata` is non-nil, `metadataChanged` is true and we fall to full reprocess, which rebuilds `tc.terminalIds` from `extractTerminalIds(items) + extractMetadataTerminalIds(tc.metadata)`.

### 9. "Revalidate unterminated fence lines" (line 948)
**Fixed.** `extendStrippedRaw` re-validates the completed/growing fence line against `isOpeningFence` and `couldBeOpeningFencePrefix`. A partial fence that grows into an invalid tag form (e.g. ` ```{.swift} `) signals `needsFullReprocess` so the line is kept in the body. `previousRawWasPartialFence` also uses `couldBeOpeningFencePrefix`, so inline-code text (e.g. ` \`cmd\` ran `) does not trigger a false positive.

### 10. "Skip replacing identical tool-call snapshots" (line 683)
**Fixed.** The `return` is guarded by all the checks above. An identical snapshot with no rawOutput/metadata change is a true no-op.

---

All tests pass (5835 tests, 8 pre-existing SSH integration failures unrelated to this change). CI is green. Could a fresh `@codex` review confirm these are resolved?

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #858, opened 2026-07-20
feat(gg): stacked-diffs phase 3 — agent integration and hardening

mrmans0n:
## Summary

Phase 3 of the gg (git-gud) stacked-diffs integration: **agent integration** and a **hardening batch**. Builds on phase 1 (detect/visualize, #850) and phase 2 (actions, #854/#855).

### Agent integration
- **Stack context in the ACP preamble** — new sessions get a stack-status section (stack name, entry count) and, when the git-gud MCP is actually attached, a note that gg tools are available. Suppressed in CLI mode and for remote projects.
- **Auto-attach `git-gud` MCP** — when gg is enabled for a local project and the repo has gg config, the `gg-mcp` stdio server (env `GG_REPO_PATH`) is composed alongside the built-in Alas MCP. The binary is resolved via `which` at availability-probe time and started on demand by the ACP layer — it is never executed during session creation.

### Create-as-stack
- The new-worktree dialog gains a **"Create as gg stack"** toggle (shown only when gg is enabled and `defaults.branch_username` is configured). Alas creates the worktree with a gg-convention `<username>/<stack-name>` branch, with a live preview.

### Hardening
- `GGServiceError.map` consolidates exit-code→error mapping across all gg invocations.
- Sync/land **completion summaries** ("Synced · N pushed · M PRs created", "Landed N PRs") surface in the drawer.
- `GGService.land` tolerates post-land JSON envelope drift **without** swallowing in-band command errors.
- Drawer gains a **refresh button**, a completion banner, and full keyboard/accessibility parity with the review-loop drawer.
- Stale stack badges are **pruned** when their worktrees disappear.

## Testing
- 42 gg-focused tests across 6 suites pass locally; full `xcodebuild` build clean.
- Full suite runs on CI.

## Notes
- `gg inbox` triage view is deferred to phase 4.

PR #854, opened 2026-07-19
feat(gg): Add GG stack actions and drawer

mrmans0n:
## Summary
- add GG stack action models and service operations for sync, land, clean, continue, abort, and checkout
- add stack drawer UI with action state, progress rows, land confirmation, and paused-operation controls
- add tests for GG action events, streaming command behavior, stack readiness, gate behavior, and right pane stack interactions

## Validation
- xcodegen
- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build\n- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test (interrupted after the process stayed silent and idle for several minutes)

PR #850, opened 2026-07-19
feat(gg): add optional git-gud stacked diffs integration (phase 1)

mrmans0n:
## Summary
- Detects gg stacks (four-gate chain: master toggle → gg installed → per-project Off/Auto/On → branch is stack-shaped) and stays fully invisible/inert without gg
- Visualizes stacks in the Commits section (PR/MR chips, CI dots, dimmed merged entries) and a sidebar `▲ merged/total` badge
- Adds Settings → Changes → "Stacked diffs" with a Homebrew install flow and per-project mode picker
- gg stack loading runs off the critical refresh path (fire-and-forget, cancellable) and re-gates reactively when settings change

## Test plan
- [x] `xcodegen && xcodebuild ... build` — clean, 0 errors
- [x] All 7 new gg suites (29 tests) pass: GGStackModelsTests, GGServiceTests, GGConfigCodableTests, GGStackGateTests, GGStackChipModelTests, GGInstallControllerTests, RightPaneGGStackTests
- [x] Existing AppConfig/ProjectConfig codable suites pass — no regression
- [x] Per-task spec+quality review, plus a whole-branch review (no Critical/security issues; two Important findings fixed and re-reviewed clean)
- [ ] CI full suite

PR #860, opened 2026-07-20
feat(gg): stacked-diffs phase 4 — inbox triage tab and hardening

mrmans0n:
## Summary

Phase 4 of the gg (git-gud) stacked-diffs integration: a project-scoped **gg Inbox** triage tab, plus two hardening fixes carried from the phase-3 ledger. Builds on phases 1–3 (#850, #854/#855, #858).

### gg Inbox tab
- A new center-pane **Inbox** tab, opened per project from the sidebar project context menu ("gg Inbox") or the stack drawer, backed by `gg inbox --json`.
- Runs `gg inbox` once per **project** (at the repo root) — one forge round-trip per refresh, never per worktree.
- Renders gg's six priority buckets (Ready to land → Draft) with PR/CI state, plus a `stack_errors` footer. Rows navigate to the matching live worktree (only the current user's `<username>/<stack>` branch) and link to the PR; rows without a live worktree are dimmed but keep the PR link.
- Refresh on open when the snapshot is missing or stale (120s), plus a manual refresh button; stale snapshots are retained across refresh failures alongside the error.

### Hardening
- `runGGSync` no longer surfaces a success summary when the sync stream reported an error.
- The new-worktree dialog seeds the base picker synchronously on project switch, so it never briefly shows the previous project's base.

## Testing
- 81 gg-focused tests across 9 suites pass locally (decode, service, store, resolver/helpers, tab dedupe, host targeting, sync summary, dialog); full `xcodebuild` build clean.
- Full suite runs on CI.

## Notes
- The inbox is navigation-first; landing/syncing stays in the per-worktree drawer.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -618,6 +618,7 @@
 		763947308C99E2D3CB565889 /* AgentKind.swift in Sources */ = {isa = PBXBuildFile; fileRef = 53C792D59DBFFF6C28F2D426 /* AgentKind.swift */; };
 		771CD72B94194370E783F7DD /* StartupScriptResolverTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9D3C76F2F0F6BCE4023A9D7F /* StartupScriptResolverTests.swift */; };
 		77CDD88F7C517095DE580B50 /* ReviewEvidence.swift in Sources */ = {isa = PBXBuildFile; fileRef = FC11F5C6667A6ECB74E066AE /* ReviewEvidence.swift */; };
+		77D7244E04B1949523699AEB /* GGInboxTabsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 649D5CB480CE056B1C4457D6 /* GGInboxTabsTests.swift */; };
 		78D25A04EAE26A0F3021A057 /* Paths.swift in Sources */ = {isa = PBXBuildFile; fileRef = 364213292935A05F3B52F51F /* Paths.swift */; };
 		78D49DF401553D2A9BCE1EE5 /* RemoteTerminalScriptTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2E14E781F0E1127387A06907 /* RemoteTerminalScriptTests.swift */; };
 		78D7FFD42803C098160BADC6 /* ProjectMCPServer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 81E981C10B13C43CE65DA74D /* ProjectMCPServer.swift */; };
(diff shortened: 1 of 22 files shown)


mrmans0n:
@codex review

mrmans0n:
Both addressed in 5762df25: the inbox now hosts and targets navigation on `visibleWorktrees(projectId:)` only.

- `AppState.openGGInbox` picks the host worktree from the visible list, so the inbox tab never lands on an archived/hidden strip (which the center-pane resolver ignores).
- `GGInboxTabView.resolveWorktreeId` resolves against visible worktrees, so a stack that only maps to an archived worktree is correctly dimmed as "No live worktree" instead of selecting a hidden id and emptying the center pane.

The pure helpers (`inboxHostWorktreeId`, `GGInboxWorktreeResolver`) are unchanged and still pass; this is a call-site swap to the visible-only list the rest of AppState already uses for selection.

@codex review

mrmans0n:
Fixed in d9e548a2. `ciIconName` now decodes through the existing `GGCIStatus` enum (whose raw values match gg's `ci_status_str` exactly), so a passing check (`"success"`) renders a checkmark and `canceled` renders an x — previously only the non-existent `"passed"` value was recognized, so successful CI showed no icon. Also added a `ciIconColorToken` helper aligned with `GGStackChip`'s mapping (success→add, failed/canceled→del, running/pending→caution) and covered both with unit tests.

@codex review

mrmans0n:
Good catch, and confirmed against gg's source: `infer_stack_usernames` (gg-core/src/commands/inbox.rs) enumerates every locally-inferred username, and `InboxEntryJson` carries only `stack_name` — no owning username or branch — so entries for two different users' same-named stacks are indistinguishable from Alas's side.

Fixed in 99859207: added `GGInboxWorktreeResolver.ambiguousStackNames(in:)`, which flags any `stack_name` that appears more than once across the full inbox snapshot. `resolveWorktreeId` now dims a row's navigation whenever its stack name is ambiguous, rather than guessing it belongs to the current user. This is deliberately conservative — even the current user's own entry gets dimmed in that rare case — since gg's JSON doesn't carry enough information to disambiguate safely.

@codex review

mrmans0n:
Two notes on the latest round:

**"Avoid routing ambiguous inbox rows"** — this was already addressed in 99859207 (`GGInboxWorktreeResolver.ambiguousStackNames`, gating `resolveWorktreeId`); the comment text is identical to the earlier round, and the line number has just drifted with the diff. The guard is in place and covers exactly the alice/auth + bob/auth scenario described.

**"Include stack errors in ambiguity detection"** — good catch, this was a real gap in that fix: `stack_errors` entries (a per-user stack that failed to load) weren't counted, so a stack that errored for one user while another user has a valid same-named entry could still bypass the guard. Fixed in 256c29f4 — `ambiguousStackNames` now takes `stackErrors` and counts them alongside bucket entries.

@codex review

mrmans0n:
Round 5 addressed in 4250b3d0 — this one exposed a real design flaw in the previous ambiguity fix, thank you:

**"Do not mark multi-PR stacks ambiguous"** — correct, and a genuine regression: counting inbox JSON *rows* sharing a `stack_name` was the wrong signal, since a single stack with several positions/PRs normally produces multiple rows with the same name. Replaced that entirely with `hasAmbiguousLocalOwners`, which checks whether more than one *local worktree branch* matches the stack-name suffix (gg infers per-user ownership from local branches via `infer_stack_usernames`, so a second matching branch is the real collision signal — independent of row count). This also fully subsumes the earlier `stack_errors` concern, so that special-casing was removed too — it's now a single, simpler, structurally-correct check folded directly into `worktreeId`.

**"Avoid routing ambiguous inbox rows" (re-posted)** — covered by the same redesign.

**"Gate refreshes from persisted inbox tabs"** — fixed: `refresh()` now checks `state.ggInboxAvailable(projectId:)` before invoking `gg inbox`, so a stale/restored tab whose project has had stacked-diffs disabled, gg uninstalled, or turned remote no longer shells out.

@codex review

mrmans0n:
Fixed in c1907573. `hasSuffix("/<stackName>")` was a raw suffix match on the whole branch string, which falsely conflated a nested stack (e.g. `alice/feature/auth`) with an unrelated top-level stack of a shorter name (`auth`), since gg stack names can themselves contain slashes. Replaced both suffix checks with a proper first-slash split (`<username>/<stackName>`, comparing only the remainder after the FIRST slash to the target name) — the exact-match path was already correct and untouched.

@codex review

PR #864, opened 2026-07-20
style(diff): simplify context expansion controls

mrmans0n:
## Summary

- remove the accent fill from expand-context controls while they are idle
- retain the existing accent feedback on hover and press
- use double directional chevrons for "Expand all context" while keeping the labels and existing hit targets
- select and measure the chevron from the action mode so both single- and double-chevron controls remain aligned

## Testing

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/DiffPaneTextDocumentBuilderTests test`
- GitHub Actions `build-test`

Part of stack `expand-button-ui`

Commit: aa53da8


mrmans0n:
@codex review

PR #866, opened 2026-07-20
fix(diff): lighten review stream row identity

mrmans0n:
## Summary

This reduces main-thread work in the diff review surface while review rows are streaming or invalidating. The ANR sample pointed at SwiftUI spending time rebuilding the lazy stack/ForEach subtree, so this narrows the top-level `LazyVStack` identity data to lightweight row descriptors instead of enumerating the full render-eligible file models directly.

## What changed

- Adds `DiffReviewRenderRow`, a small stable row descriptor containing only the file index, file identity, and bottom-spacing flag.
- Builds render rows from the existing render-eligible file IDs before entering the top-level `ForEach`.
- Keeps the heavy `DiffReviewFile` lookup inside each row body, reducing the amount of data SwiftUI has to diff at the stack level.
- Replaces the conditional bottom spacer with a stable zero-height/14pt `Color.clear` spacer so row shape does not churn as the last row changes.
- Adds focused tests for empty render-row output and stable index/id/spacing behavior.

## Verification

- `xcrun swiftc -parse -parse-as-library` on the changed Swift files
- `git diff --check`
- GitHub Actions `Build` workflow

Part of stack `moar-beachbooooooools`

Commit: e58428f

### DIFF
--- a/Alas/Sources/Center/DiffReview/DiffReviewRenderEligibility.swift
+++ b/Alas/Sources/Center/DiffReview/DiffReviewRenderEligibility.swift
@@ -1,9 +1,26 @@
 import Foundation
 
+struct DiffReviewRenderRow: Identifiable, Equatable {
+    let index: Int
+    let id: DiffReviewFileID
+    let showsBottomSpacing: Bool
+}
+
 /// Keeps the review stream deterministic: every loaded file is render-eligible,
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #881, opened 2026-07-21
feat(open-binary): graceful binary file handling and breadcrumb context menus

mrmans0n:
## Summary

Open binary files (PDFs, videos, archives, executables, etc.) with a dedicated preview tab offering an "Open with System" affordance, instead of falling through to the text editor and rendering a stale read-only "file is not valid UTF-8" message. Adds system-open/reveal-in-Finder actions to the image preview pane, the editor tab-bar context menu, and right-click context menus on every breadcrumb segment.

## Changes

**New files**
- `BinaryFileType` — known-binary extension set (pdf, mp4, zip, exe, dylib, …), mirrors `ImageFileType`.
- `FileSystemOpen` — thin `NSWorkspace` wrapper (`open` / `reveal`).
- `BinaryPreviewTabView` — placeholder view: icon + "Binary file" + Open with System + Reveal in Finder.
- `BreadcrumbView` — shared breadcrumb with per-segment right-click context menus (file vs folder), replaces the inline breadcrumbs in `EditorTabView` and `ImagePreviewTabView`.

**Routing**
- `AppState.openFile` routes known-binary extensions to `openBinaryPreview` (between the image and editor branches), skipping a wasted text-load attempt.
- The external-file path (`alas open` on a path outside any worktree) gets the same binary short-circuit.
- Unknown-extension binaries still reach the editor, which now exposes an observable `loadKind` on `EditorBuffer`; when it reports `.notUTF8`, `EditorTabView` renders the same binary placeholder overlay instead of `CodeEditorView`.

**Affordances**
- `ImagePreviewTabView` footer gains "Open with System" + "Reveal in Finder" buttons; the missing/unsupported/decode-failed placeholders get an "Open with System" button.
- The per-tab context menu in `TabBarView` gains "Open with System" + "Reveal in Finder" (gated on `tab.relativeFilePath != nil`).
- File-name breadcrumb segment (the last segment): View at HEAD, Compare with HEAD, File History, Copy Relative Path, Copy Full Path, Reveal in Finder, Open with System. Git-specific items are hidden for external editor tabs.
- Intermediate-folder breadcrumb segment: Reveal in Finder, Focus on Files tab (switches the right pane to Files and expands to that folder via the existing `revealInFiles`/`rps.reveal(path:)` mechanism), Copy Full Path.

## Tests

- `BinaryFileTypeTests` — extension matching (known/unknown/image/empty/no-extension).
- `OpenBinaryRoutingTests` — `openBinaryPreview` find-or-create (reuse, new tab, activation); `AppState.openFile` routing (binary → `.binaryPreview`, image → `.imagePreview`, text → `.editor`).
- Existing `EditorBufferTests` (74) and the full suite (6063) still pass. The only failing tests are `SSHIntegrationTests` (8), which require `ALAS_SSH_INTEGRATION=1` and are unrelated/environmental.

## Out of scope

Hex viewer, per-extension custom handlers, in-place binary editing, "Open with…" picker.


mrmans0n:
Addressed all three P2 items in `84f8eb0f`:

1. **Absolute breadcrumb prefixes** — `BreadcrumbView` now preserves the leading slash for absolute paths, so folder-menu actions on `/tmp/a.zip` resolve against `/tmp` instead of `tmp`.
2. **External binary paths in tab bar** — added `Tab.absoluteFilePath` (returns the absolute path for external editor tabs and external binary previews, nil otherwise). `onOpenWithSystem` / `onRevealInFinder` use it as a fallback before `relativeFilePath`. `Tab.relativeFilePath` for `.binaryPreview` now returns nil when the path is absolute (mirroring `.editor`'s `isExternal` guard), so the existing `Copy Path`/`Copy Relative Path` handlers correctly skip external binary tabs instead of appending the absolute path to `worktree.path`.
3. **Remote `loadKind` reset** — both successful remote `.file` paths (restored-snapshot and fresh-load) now set `loadKind = .loaded`, so a remote file that transitions from binary/unreadable to valid UTF-8 will correctly swap the placeholder back to the editor.

mrmans0n:
Addressed all four round-2 P2 items in `ed557468`:

1. **`.dat`/`.pak` removed from known-binary list** — they're ambiguous (some repos use UTF-8 `.dat` fixtures). `.bin` kept (conventionally binary). Files with these extensions now fall through to the editor, which still detects `.notUTF8` via the content sniff and shows the binary placeholder if appropriate.
2. **Finder/system-open gated for remote worktrees (BinaryPreviewTabView)** — added `isRemote = worktreePath.isRemoteAlasPath`; the placeholder buttons and the breadcrumb Reveal/Open menu items are now `nil` (omitted) for remote worktrees, matching the `ChangesTabView` pattern.
3. **`loadKind` reset on restore** — `finishInitialLoad` now tracks whether a hot-exit snapshot or pending user edits were applied and resets `loadKind = .loaded` in that case, so `EditorTabView` keeps showing the editor with the user's recoverable draft instead of swapping to the binary placeholder.
4. **Finder/system-open gated for remote worktrees (EditorTabView)** — same `worktreePath.isRemoteAlasPath` guard applied to the breadcrumb menu items and the binary placeholder buttons. Also applied to the tab-bar `onOpenWithSystem`/`onRevealInFinder` handlers in `CenterPaneView`.

mrmans0n:
@codex review

mrmans0n:
Addressed the round-3 P3 in `d8532974`: tap-to-reveal-in-Files and the folder-menu "Focus on Files tab" action are now suppressed for absolute (external) breadcrumb paths — `onRevealInFiles` treats paths as worktree-relative, so feeding `/tmp/...` would have expanded a bogus node.

mrmans0n:
Addressed both round-4 P2 items in `82afefe6`:

1. **Clear `readOnly` on failed-load restore** — `finishInitialLoad` now clears `readOnly = false` (alongside `loadKind = .loaded`) when content was restored AND the on-disk load failed (`loadKind != .loaded`). This avoids the symlink test regression: symlinks/external files that loaded successfully stay read-only even when the user typed before the load completed, but binary/unreadable/missing files with a recoverable draft become editable.
2. **External editor tabs use `peekExternalBuffer` for `isBinary`** — added `TabsManager.peekExternalBuffer(tabId:)` (backed by `EditorBufferStore.peekExternalBuffer`, a non-creating lookup) and wired `EditorTabView.isBinary` to use it for external tabs. External binaries that went through the editor path (unknown extension + external, or known binary opened with a line target) now correctly show the binary placeholder.

mrmans0n:
Addressed the round-5 P2 in `de6dc0ca`: added `Tab.supportsSystemOpenActions` (true for `.editor`, `.imagePreview`, `.binaryPreview`, `.mergeConflict`; false for `.stashDiff`, `.fileSnapshot`, `.fileHistory` whose `relativeFilePath` points at a historical/stashed revision). The tab-bar "Open with System" / "Reveal in Finder" items are now gated on it, so they only appear for tabs representing the live, current on-disk file. "Copy Path" / "Copy Relative Path" stay gated on `relativeFilePath != nil` (copying the path is informational and harmless).

mrmans0n:
@codex review

mrmans0n:
Addressed both round-6 P3 items in `0fa7033d`:

1. **Hide system actions for remote tab bar** — added `systemActionsEnabled: Bool` to `TabBarView` (defaults true), wired from `CenterPaneView` as `!worktree.path.isRemoteAlasPath`. The "Open with System" / "Reveal in Finder" menu items now hide for remote worktrees instead of rendering enabled-but-no-op.
2. **Drop "Open with System" for `.missing` images** — the `.missing` placeholder no longer passes an `onOpenWithSystem` handler (the file doesn't exist, so the button was a dead action). `.decodeFailed` and `.unsupported` still offer it (the file exists but can't be previewed in-app).

PR #887, opened 2026-07-22
fix(gitlab): send Content-Type header on glab api --input requests

mrmans0n:
## Summary

Publishing a review from the PR/MR review pane to GitLab failed on every draft comment with `glab api merge request discussions failed: glab: HTTP 415`.

Root cause: `glab api` does **not** set a `Content-Type` header when the request body is read from `--input` (verified with `GLAB_DEBUG_HTTP=1`), so GitLab rejects the JSON body with HTTP 415 Unsupported Media Type. The older `-f` form-field path sets the header automatically, which is why other calls worked.

Fix: pass an explicit `-H "Content-Type: application/json"` on all five `--input -` call sites in `GitLabCLIProvider` (create discussion, approve, MR note, discussion note, resolve).

## Testing

- `GitLabCLIProviderTests` (88) and `GitLabCLIProviderVerdictTests` (5) pass.
- Updated exact-arg assertions in `threadMutationsUseDiscussionEndpointsAndRefreshMR` and added a `sendsJSONContentType` assertion to the publish-review test to lock in the header.

### DIFF
--- a/Alas/Sources/Integrations/CodeHost/GitLabCLIProvider.swift
+++ b/Alas/Sources/Integrations/CodeHost/GitLabCLIProvider.swift
@@ -880,6 +880,9 @@ struct GitLabCLIProvider: CodeHostProvider {
                 Self.mergeRequestAPIPath(remote: remote, request: request, suffix: "discussions"),
                 "--method", "POST",
                 "--hostname", remote.host,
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #888, opened 2026-07-22
fix(diff): new diff comment composers

mrmans0n:
## Summary

This fixes a regression where clicking the gutter add-comment button could render a new diff draft composer without immediately moving keyboard focus into its text field. The focus request is now tied to the user action that opens or relocates the composer, rather than to SwiftUI appearance timing.

### Durable Composer Focus

- Added an explicit `focusRequestGeneration` path to `ReviewDraftComposerTextEditor` so each new comment action can request AppKit focus exactly once after the text view is attached to a window.
- Preserved the legacy `FocusState` behavior for existing callers that do not pass an explicit generation.
- Cancel stale scheduled focus work when the representable is dismantled or superseded.

### Diff Surface Wiring

- Routed gutter comment actions in both `DiffReviewFileSection` and `DiffTabView` through a shared `beginPendingDraft` helper.
- Increment the focus generation whenever a draft is started, including when the composer moves between different selected diff lines.
- Clear composer focus state when the draft is cancelled or saved.

### Regression Coverage

- Added a direct harness test proving the shared AppKit text editor refocuses for each new focus generation.
- Added end-to-end surface tests for the file review surface and diff tab surface that select different gutter rows and assert the new composer owns first responder each time.

## Where to start

- `Alas/Sources/Center/DiffReview/DiffReviewFileSection.swift` - shared composer focus behavior and file review wiring
- `Alas/Sources/Center/DiffTabView.swift` - diff tab gutter comment wiring
- `AlasTests/DiffReviewSurfaceTests.swift` - focused regression coverage for the shared editor and review surface

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/DiffReviewSurfaceTests -only-testing:AlasTests/DiffPaneViewTests -quiet`

Full `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` ran 6305 tests and failed with unrelated existing issues: 8 `SSHIntegrationTests` require `ALAS_SSH_INTEGRATION=1`, and 2 `RightPaneGGStackTests` manual rebase expectation failures reproduce when run in isolation.

### DIFF
--- a/Alas/Sources/Center/DiffReview/DiffReviewFileSection.swift
+++ b/Alas/Sources/Center/DiffReview/DiffReviewFileSection.swift
@@ -118,6 +118,7 @@ struct DiffReviewFileSection: View {
     @StateObject private var renderContextCache = DiffReviewRenderContextCache()
     @State private var pendingDraftAnchor: DiffReviewLineAnchor?
     @State private var pendingDraftBody = ""
+    @State private var draftComposerFocusRequestGeneration = 0
     @State private var expandedCollapsedRowIDs: Set<String> = []
     @State private var contextSnapshot: DiffReviewFileContextSnapshot?
     @State private var contextExpansion = DiffContextExpansionState()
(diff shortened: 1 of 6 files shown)


mrmans0n:
@codex review

PR #896, opened 2026-07-22
feat(gg): add mode selection to new worktree creation

mrmans0n:
## Summary

- replace the new-worktree GG Boolean with an Inherit / On / Off mode picker
- share linked-worktree GG policy resolution between creation and runtime gating
- stage optimistic overrides in runtime state and persist explicit modes only after successful reconciliation
- preserve the selected raw mode when retrying a failed worktree creation

## Why

The previous “Create as gg stack” toggle controlled branch and base creation but did not persist a worktree policy. A new linked worktree therefore inherited the repository default, which could activate GG even when the creation toggle was left off. The new control maps directly to the persisted worktree policy while keeping inherited behavior explicit.

## Validation

- `xcodegen`
- focused Swift Testing coverage for policy resolution, optimistic persistence, failure/retry behavior, dialog state, and preview formatting

### DIFF
--- a/Alas/Sources/App/AppState.swift
+++ b/Alas/Sources/App/AppState.swift
@@ -1252,7 +1252,11 @@ final class AppState {
                     )
                     projectsManager.setOperationState(
                         id: optimistic.id,
-                        state: .createFailed(message: error.localizedDescription, base: base)
+                        state: .createFailed(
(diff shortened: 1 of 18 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #897, opened 2026-07-23
fix(acp): measure transcript text height independent of view bounds

mrmans0n:
Reopened transcripts still drew short rows on top of the next one — user
messages and single-line markdown blocks overlapping the paragraph below.

Root cause: ACPMarkdownInlineNSTextView measured height via
NSLayoutManager.usedRect after setting textContainer.containerSize, but the
container has widthTracksTextView = true, so it ignores that width and
tracks the view's own bounds width instead. When SwiftUI probes
sizeThatFits before the row has its final frame, bounds width is stale or
zero, so the text measured as a single unwrapped line. The height cache
added in #889 then froze that too-small height (before #889 every scroll
frame re-measured and self-corrected, at the cost of the beachball #889
fixed), so SwiftUI laid the row out short and the next row drew over it.

Measure the attributed content itself with
NSAttributedString.boundingRect(with:options:) — a pure function of the
text and the width, independent of the view's bounds, TextKit version, and
widthTracksTextView. Drawing still wraps at the frame width SwiftUI assigns
from that measurement, so draw matches measure.

Add regression tests asserting a freshly-populated view's first probe
reports the real multi-line height and agrees with an independent bounding
measurement. Drop #889's two container-width-restoration tests, which
pinned the replaced usedRect approach.


mrmans0n:
@codex review

PR #903, opened 2026-07-23
fix(terminal): preserve persistent tabs across Alas instances

mrmans0n:
## Summary

- prevent a second Alas instance launched from a persistent terminal from inheriting the parent zmx session identity
- explicitly shadow `ZMX_SESSION` only for local Ghostty surfaces so zmx performs a normal multi-client attach
- preserve existing remote SSH behavior and `ZMX_SESSION_PREFIX`
- add direct environment and Ghostty C-boundary regressions for both local and remote terminals

## Root cause

Ghostty starts terminal processes with the Alas process environment and then overlays the environment supplied by `EnvBuilder`. Omitting inherited `ZMX_SESSION` from the overlay did not remove it when Alas itself was launched inside a zmx-backed terminal. zmx therefore interpreted the restored terminal's attach as a session switch, causing the original attach client to exit and Alas to close and kill the shared persisted terminal.

## Validation

- `xcodegen`
- 52 focused tests across `EnvBuilderTests`, `GhosttyConfigBuilderTests`, `RemoteTerminalLaunchTests`, and `TerminalServiceZmxTests`
- test-first RED/GREEN coverage for the local attach bug and remote-scope guard

### DIFF
--- a/Alas/Sources/Terminal/EnvBuilder.swift
+++ b/Alas/Sources/Terminal/EnvBuilder.swift
@@ -4,16 +4,8 @@ enum EnvBuilder {
     private static let strippedKeys: Set<String> = [
         "TERM", "TERMINFO", "TERMINFO_DIRS",
         "TERM_PROGRAM", "TERM_PROGRAM_VERSION", "COLORTERM",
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #906, opened 2026-07-23
fix(sidebar): mute pending stack indicator

mrmans0n:
## Summary

- Mute the pending gg stack sidebar indicator before any synced stack summary exists.
- Add regression coverage for the pending stack indicator color token.

## Root cause

The left sidebar used the `accent` theme token whenever gg was active but the worktree did not yet have a `GGStackSummary`. That made an unsynced/no-commit-synced stack look accented before there was synced stack state to report.

## Validation

- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet test -only-testing:AlasTests/WorktreeRowHeightTests`
- `xcodegen`

Full `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` was attempted locally but hung silently for about 17 minutes and was interrupted.

### DIFF
--- a/Alas/Sources/Sidebar/WorktreeRowView.swift
+++ b/Alas/Sources/Sidebar/WorktreeRowView.swift
@@ -52,6 +52,10 @@ struct WorktreeRowView: View {
         stackSummaryText(merged: merged, total: total)
     }
 
+    static func pendingStackIndicatorColorToken() -> String {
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #917, opened 2026-07-24
fix(gg): new worktree name field initialization

mrmans0n:
## Summary

- Initialize the New Worktree dialog's selected project before the first SwiftUI render.
- Keep regular branch and GG stack-name drafts independent.
- Add unit coverage for initial project selection with valid, stale, absent, and empty project inputs.

## Root Cause

The dialog previously started with an empty `projectId`, so the focused name field initially bound to the regular branch draft. `.onAppear` then selected the first or preset project; when that project inherited GG mode, the same visible field switched to the stack-name draft. Early keystrokes could land in the regular branch draft and disappear from the visible stack-name field.

## Validation

- `rtk xcodegen`


mrmans0n:
👍 Looks good. CI is green, the fix is scoped to initial project selection, and the tests cover the branch/stack-name initialization behavior.

PR #926, opened 2026-07-28
fix(gg): preserve first worktree name keystroke

mrmans0n:
## Summary

- resolve the initial repository and its effective Stacked Diffs Mode together before the New Worktree dialog first renders
- seed both `projectId` and `ggMode` from that atomic selection so immediate input targets the correct branch or stack-name draft
- preserve independent branch and stack-name drafts when toggling mode manually
- add focused regression coverage for enabled, disabled, preset, stale-preset, and empty-project initialization

## Root cause

The dialog resolved its initial repository before the first render, but `ggMode` still started as `.off` and changed during `onAppear`. A fast first keystroke could therefore be stored in the regular branch draft immediately before the field switched to the independent stack-name draft, making the character appear to be dropped.

## Verification

- `rtk swiftformat Alas AlasTests --lint --reporter github-actions-log`
- `rtk xcodegen`
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -derivedDataPath /tmp/alas-stacked-diffs-input-green -only-testing:AlasTests/NewWorktreeDialogTests test`
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -derivedDataPath /tmp/alas-stacked-diffs-input-build -quiet build`
- `rtk git diff --check`

The broad local test invocation compiled but stalled before starting `xctest`, so GitHub Actions remains the exhaustive suite gate.

### DIFF
--- a/Alas/Sources/Dialogs/NewWorktreeDialog.swift
+++ b/Alas/Sources/Dialogs/NewWorktreeDialog.swift
@@ -27,7 +27,7 @@ struct NewWorktreeDialog: View {
     @State private var branch: String = ""
     @State private var stackName: String = ""
     @State private var runStartup: Bool = true
-    @State private var ggMode: GGWorktreeMode = .off
+    @State private var ggMode: GGWorktreeMode
     @State private var openAfterCreate: Bool = true
     @State private var launchMode: AppConfig.LauncherMode = .terminal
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #928, opened 2026-07-28
fix(diff): Keep diff word wrap pane-local and off by default

mrmans0n:
## Summary

- make word wrapping transient and pane-local across every diff surface that exposes the toggle
- initialize each new diff pane with word wrap off and preserve distinct state when switching between diff tabs
- retire the persisted `diffWrapLines` setting while continuing to decode older config files safely
- add regression coverage for local bindings, persistence side effects, tab identity, and legacy config migration

## Validation

- `xcodegen`
- `swiftformat Alas AlasTests --lint --reporter github-actions-log`
- full macOS test suite before the clean rebase onto current `origin/main`
- focused diff preference/tab identity tests on the rebased head
- quiet macOS build on the rebased head

### DIFF
--- a/Alas/Sources/Center/CenterPaneView.swift
+++ b/Alas/Sources/Center/CenterPaneView.swift
@@ -231,6 +231,7 @@ struct CenterPaneView: View {
                                 rps.requestDiscardFile(path: s.relativePath)
                             }
                         )
+                        .id(s.id)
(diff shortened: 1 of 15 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #934, opened 2026-07-30
fix(acp): fix slash-command autocomplete rendering and missed suggestions

mrmans0n:
## Summary
- Long usage hints (e.g. `release-manager`'s bracketed argument list) forced their full intrinsic width via `fixedSize`, overflowing the picker panel instead of truncating with an ellipsis.
- `promptSuggestions` arriving after the user already typed `/` (e.g. right after a session takeover re-attaches) never re-opened the panel, since the panel only reconciled on keystrokes — now it also reconciles when suggestions transition from empty to populated.

## Test plan
- [x] Built and ran the app; typed `/release-manager` and confirmed the hint truncates instead of overflowing the panel
- [x] Reproduced the missed-panel case with a session whose suggestions arrived after typing `/`, confirmed the panel now opens

### DIFF
--- a/Alas/Sources/ACP/UI/ACPComposer.swift
+++ b/Alas/Sources/ACP/UI/ACPComposer.swift
@@ -84,12 +84,12 @@ struct ACPInputField: NSViewRepresentable {
 
     func updateNSView(_ nsView: NSScrollView, context: Context) {
         context.coordinator.isFocused = $isFocused
-        // The agent can send `available_commands_update` after the user has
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #935, opened 2026-07-30
fix(gg): preserve input across availability probe

mrmans0n:
## Summary

- preserve the first New Worktree name character when GG availability resolves after the dialog appears
- transfer only startup-probe input into an empty stack draft while preserving independent manual drafts
- add focused regression coverage

## Root cause

PR #926 seeded the project and Stacked Diffs Mode before the first render, but `GGAvailability.shared.isInstalled` still begins as `false` until the startup probe completes. Input could land in the regular branch draft before availability switched the field to the empty stack draft.

## Verification

- `rtk swiftformat Alas AlasTests --lint --reporter github-actions-log`
- `rtk xcodegen`
- macOS app build with isolated DerivedData
- focused `NewWorktreeDialogTests`
- adjacent `GGAvailabilityTests`, `GGStackCreateModeTests`, and `GGWorktreeContextTests`
- `rtk git diff --check`

The broad local test suite was attempted, but the build stalled before spawning `xctest`; GitHub Actions remains the exhaustive suite gate.

### DIFF
--- a/Alas/Sources/Dialogs/NewWorktreeDialog.swift
+++ b/Alas/Sources/Dialogs/NewWorktreeDialog.swift
@@ -192,6 +192,16 @@ struct NewWorktreeDialog: View {
                 if let pinned = stackPinnedBase { base = pinned }
             }
         }
+        .onChange(of: GGAvailability.shared.isInstalled) { wasInstalled, isInstalled in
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #936, opened 2026-07-30
perf(diff): bound review render tree

mrmans0n:
## Summary
- Cap automatic diff-review rendering across the loaded session instead of only per file.
- Defer files beyond the cap through the existing placeholder and explicit Show full diff action.
- Skip redundant active-comment highlight redraws in diff text panes.

## Root cause
The review stream retained every materialized LazyVStack section, allowing the SwiftUI layout tree to grow without a session-wide bound.

## Validation
- xcodegen
- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build\n- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test


mrmans0n:
@codex review

PR #937, opened 2026-07-31
fix(markdown): stop external buffer observer loop

mrmans0n:
## What changed

- Avoid rewriting external-tab observation state when the buffer mapping is unchanged.
- Start file watching only when a new external buffer instance is created.
- Add a regression test proving repeated external-buffer access does not invalidate observers.

## Why

`MarkdownTabView.body` reads its external buffer during SwiftUI rendering. `TabsManager.externalBuffer(...)` previously rewrote `@Observable` dictionaries and restarted the file watcher on every read, continuously dirtying the observation graph and leaving `flushTransactions()` unable to drain.

## Impact

External Markdown tabs no longer beachball the app through a steady-state main-thread observation loop. Editable buffer upgrades and LSP-open retries remain unchanged.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` (the regression passes; the local aggregate run retains unrelated `DiffReviewSurfaceTests` accessibility failures also reproducible outside this change)

### DIFF
--- a/Alas/Sources/Center/TabsManager.swift
+++ b/Alas/Sources/Center/TabsManager.swift
@@ -1447,23 +1447,32 @@ final class TabsManager {
         language: String? = nil,
         editable: Bool = false
     ) -> EditorBuffer {
-        externalTabURLs[tabId] = (worktreeId: worktreeId, url: absoluteURL)
+        let existingEntry = externalTabURLs[tabId]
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #938, opened 2026-07-31
feat(gg): stream progressive inbox refresh

mrmans0n:
## Summary
- stream gg inbox JSONL results into Inbox as each stack completes
- require gg 0.9.12+ and offer an in-app upgrade path
- preserve cached results during refresh and surface actionable refresh errors

## Testing
- xcodegen
- xcodebuild build
- focused Inbox and installer tests
- full suite: 6,710 passed; one existing flaky DiffReviewSurface accessibility test failed twice, and its isolated rerun passed


mrmans0n:
@codex review

PR #951, opened 2026-08-04
fix(worktree): activity timestamps from HEAD commits

mrmans0n:
## Summary

- Derive local worktree activity from each worktree's checked-out HEAD commit time.
- Prevent packed Git refs from making every imported worktree appear to have the same last-update time.
- Preserve existing local folder creation timestamps and upstream SSH metadata behavior.

## Root cause

When branches were stored in Git's shared `packed-refs` file, Alas used that single file's modification time for every worktree. The sidebar and automatic last-update sort therefore could not distinguish worktrees with different commit histories.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/WorktreeServiceTests -only-testing:AlasTests/RemoteLastActivityTests` (42 tests passed)

The full local suite has an unrelated, repeatable Diff Review accessibility-test failure; GitHub Actions is the final CI gate for this PR.

### DIFF
--- a/Alas/Sources/Git/WorktreeService.swift
+++ b/Alas/Sources/Git/WorktreeService.swift
@@ -49,6 +49,8 @@ struct WorktreeService {
         )
         if let host {
             trees = await Self.fillingRemoteMetadata(trees, host: host)
+        } else {
+            trees = await Self.fillingHeadCommitTimes(trees)
         }
         return trees
     }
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

PR #960, opened 2026-08-05
fix(acp): make the tail-follow pause reachable under responsive scrolling

mrmans0n:
Rebased onto `main` (now includes #950). Scope is narrower than the original description: #950 already fixed the re-pin-during-gesture problem, so what remains here is the one gap it left open.

## The gap

#950 established that `NSApp.currentEvent` is unusable as a scroll-intent signal under responsive scrolling — it measured 1176 of 1209 ticks (97%) as not-user-driven. It removed that gate from the pagination decisions and neutralised the re-pin with `noteUserScroll()` / `repinsToTail(followsTail:wasFollowingTail:)`, which use the reliable programmatic/non-programmatic split.

It deliberately left the gate in place for the tail-follow **pause**:

```swift
let isHeadPaginationDriven = ACPUserScrollEvent.isHeadPaginationDriven(...)
decide(..., isUserDriven: isHeadPaginationDriven)
```

`ACPScrollDirectionClassifier.decide` only returns `.userScrolledUp` for a user-driven tick, so `pauseTailFollow()` fires only on the ~3% of ticks that coincide with a recognisable fresh event. A reader who scrolls up out of the tail therefore keeps `session.followsTranscriptTail` set. During the gesture that is harmless — `repinsToTail` suppresses the re-pin — but the suppression window is deliberately short-lived ("everything the window suppresses is behavior the user gets back the moment they stop"). Once the gesture ends, `repinsToTail` returns `!isUserScrollInFlight()` → `true`, and the next update pins them back to the bottom.

An independent capture on this branch measured the same signal quality: **414 of 417** ticks not-user-driven (`mouseEntered` 309x, `mouseExited` 73x, `cursorUpdate` 28x, `scrollWheel` 3x), `pauseTailFollow()` called **0** times across a gesture that reached 470pt from the bottom — well past the 160pt `pauseTolerance`.

## Fix

`ACPTranscriptScrollerView` tracks the scroll view's own `willStartLiveScroll`/`didEndLiveScroll` notifications, which AppKit posts only for genuine user scrolling and never for a programmatic `setBoundsOrigin`, and exposes `isUserScrollActive` with a grace period (matching `ACPUserScrollEvent.freshnessWindow`) so momentum and the elastic bounce still count as the gesture.

That is OR-ed into `isHeadPaginationDriven`, the pause gate only. Pagination stays geometric, exactly as #950 left it. The existing event test is kept alongside, since it covers input that never opens a live-scroll session (scrollbar-track click, magnify, swipe).

This is complementary to `noteUserScroll()` rather than a replacement: that one answers "is a gesture in flight right now" to suppress re-pinning; this one answers "was this tick the user" so the pause can latch and the state converges instead of relying on the suppression window.

## Tests

New suite `ACPTranscriptScrollerLiveScrollTests` (4 tests). `NSApp.currentEvent` is nil throughout a unit-test run, which is exactly the condition the old gate failed under:

- an upward live scroll past the pause tolerance pauses tail-follow
- a tail row re-measuring afterwards does not drag the viewport back
- a programmatic scroll is not mistaken for a user gesture
- live-scroll activity lapses after the gesture settles

111 tests across 11 scroller suites pass, including #950's `ACPTranscriptScrollerScrollBackTests`.

## Follow-ups (not in this PR)

- `headStepThreshold` is `max(1500, viewportHeight * 2)`, an absolute distance from the document top. On a transcript whose whole scroll range is smaller than that, head pagination fires even at the bottom; a capture showed a 30-row insert growing the document 2041pt mid-gesture.
- Every streaming tick classifies as `.reset` (59 in 15s, unchanged row count), each re-measuring rows into slightly different heights and wandering the document height by 16-58pt.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -1297,6 +1297,7 @@
 		DB29057E7147A0774BC2B835 /* ReviewLoopDrawer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2D4CC3ED497796474F107EBE /* ReviewLoopDrawer.swift */; };
 		DB305015BF666EEE0F9C7BAB /* ACPFileEditCard.swift in Sources */ = {isa = PBXBuildFile; fileRef = 29D7631D754795D16E946440 /* ACPFileEditCard.swift */; };
 		DB40AA7EB6A31B5F232EDCA3 /* DiffPaneTextDocumentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACCA6B39E2B172FADFD9E0E6 /* DiffPaneTextDocumentView.swift */; };
+		DB474070DDBFA87828543D60 /* ACPTranscriptScrollerLiveScrollTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F1BA8B1CD4334FF7F7082E9A /* ACPTranscriptScrollerLiveScrollTests.swift */; };
 		DBE7AD7F869B444A7C48E5B2 /* TreeSitterJava in Frameworks */ = {isa = PBXBuildFile; productRef = 33EB43DCB691861654EC147A /* TreeSitterJava */; };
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #976, opened 2026-08-06
fix(acp): smooth transcript fling rebound

mrmans0n:
## Summary
- Keep native AppKit transcript pagination grow-only during wheel and trackpad flings.
- Avoid tail-follow re-pins while AppKit settles its elastic bottom rebound.

## Root cause
The 0.14.2 logical-scrollbar change bounded both window edges during a fling, forcing a reconciler reset. At the live tail, remeasured rows could also call `scrollToBottom()` during elastic settling.

## Validation
- Manual ACP transcript fling test at both edges
- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test` (7,653 tests / 698 suites)


mrmans0n:
@codex review

mrmans0n:
Addressed the two Codex findings in b7d18a24 (settled window compaction and deferred tail re-pin). @codex review

PR #979, opened 2026-08-06
fix(gg): recover cancelled stack refreshes

mrmans0n:
## Summary

- Prevents the GG stack drawer from remaining in `Loading` when an in-flight stack refresh is cancelled.
- Restores the prior stable stack only when its key still matches the live branch and commit identity.
- Shows a retryable failure instead of stale stack data for first-load or changed-key cancellations.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination "platform=macOS" -quiet build`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination "platform=macOS" -quiet test` (7,664 passed, 12 skipped)
- Codex review: thumbs up

### DIFF
--- a/Alas/Sources/Right/RightPaneState.swift
+++ b/Alas/Sources/Right/RightPaneState.swift
@@ -983,6 +983,33 @@ final class RightPaneState: GGSplitCommitServicing {
             await reconcileGGUndoCandidateIfNeeded()
             return
         }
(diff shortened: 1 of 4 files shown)


mrmans0n:
@codex review

PR #988, opened 2026-08-07
fix(review): keep AppKit diff scroll position when first comment composer appears

mrmans0n:
## Problem

On the AppKit-based diff review panes, creating the first comment on a file caused the scroll viewport to reset toward the top of the document, scrolling the just-opened comment composer out of view.

## Root cause

When a draft comment composer first appears on a group, the group's row plan restructures: the previously fused single hunk row (e.g. one ~2200pt row with id `...:group:<id>:header`) splits into a small 37pt group header row plus per-segment rows and the composer row.

The AppKit diff scroller anchors the viewport by (row id, intra-row offset) during plan reconciliation. Because the fused hunk row and the segmented group header **share the same row id**, the anchor row was still found after the restructure, but it had shrunk from ~2200pt to 37pt. `viewportMinY(for:)` clamped the captured intra-row offset (e.g. 1148) into the new 37pt row, snapping the viewport to the top of the group header — the observed "reset to top".

Subsequent comments on the same file did not reproduce the issue because the group was already in the segmented structure, so no row-id reuse with a drastically different geometry occurred.

## Fix

In `AppKitDiffTilingController.viewportMinY(for:)`, treat an anchor whose captured intra-row offset exceeds the new row height (plus a small slack for normal measurement drift) as **unresolvable** — return `nil` instead of clamping into the shrunken row. The reconciler already preserves the absolute viewport Y when the anchor cannot be resolved (clamped by the native scroll view if the document shrank), so the viewport stays where the reviewer left it and the composer remains visible.

The slack tolerance (`anchorSlack = 24pt`) keeps normal measurement differences between the captured and applied plans from being misclassified as a restructure, preserving the existing anchor-restoration behavior for genuine height drift.

## Tests

- `AppKitDiffTilingControllerTests`: anchor restoration now reports `nil` for restructured (shrunken) anchor rows; tolerates small drift; clamps negative offsets.
- `AppKitDiffScrollerReconcilerTests`: a restructured anchor row keeps the previous absolute viewport position.
- `DiffReviewSurfaceTests.appKitReviewKeepsViewportWhenFirstComposerAppearsBelowFold`: end-to-end reproduction — scroll below the fold, select a review line, assert the composer mounts and the viewport stays within 250pt of its prior position. Fails before the fix (1117pt jump), passes after.

### DIFF
--- a/Alas/Sources/Center/Diff/Scroller/AppKitDiffRowSpec.swift
+++ b/Alas/Sources/Center/Diff/Scroller/AppKitDiffRowSpec.swift
@@ -31,6 +31,7 @@ struct AppKitDiffRowSpec {
     let id: String
     let ownerID: String?
     let equalityToken: AppKitDiffRowEqualityToken
+    let contentSignature: Int
     let estimatedHeight: CGFloat
     var retention: AppKitDiffRowRetention = .recyclable
     var contextRowCount: Int? = nil
--- a/Alas/Sources/Center/Diff/Scroller/AppKitDiffScrollView.swift
+++ b/Alas/Sources/Center/Diff/Scroller/AppKitDiffScrollView.swift
@@ -137,6 +137,19 @@ final class AppKitDiffScrollView: NSScrollView {
         }
     }
 
+    /// Sets the scroll position without clamping to the document height.
+    /// Used during plan restructuring when the document height is still
+    /// based on estimates and the original viewport position may be beyond
+    /// the estimated document — clamping would move the viewport to the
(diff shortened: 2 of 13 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #990, opened 2026-08-07
fix(follow): Anchor follow revision popover

mrmans0n:
## What changed

- Anchor the follow-revision editor popover to the triggering button or menu.

## Why

The popover was attached to the full-width revision control, so it pointed to the center pane instead of the control that opened it.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build` *(blocked: GhosttyKit.xcframework is absent in this worktree)*


mrmans0n:
@codex review

PR #980, opened 2026-08-06
feat(review): migrate diff panes to AppKit scroller

mrmans0n:
## Summary
- add a feature-flagged AppKit virtualized scroller for diff and review panes
- preserve review actions, context expansion, image feedback, drafts, and GG split-preview behavior
- add recycling, scrolling, stress, and regression coverage

## Verification
- `rtk xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test`

PR #991, opened 2026-08-07
fix(review): snap file-click scroll instead of always animating

mrmans0n:
## Summary

- Clicking a file in the review rail always animated the diff pane's scroll into view, which got old fast on repeated clicks.
- The AppKit scroller (default path since #980) now snaps instantly for jumps longer than one viewport, and only keeps the slide animation when the destination is already close by.
- The legacy SwiftUI fallback path (used when the AppKit scroller flag is off) drops the animation entirely, since it has no cheap way to measure the current scroll offset for a distance check.

### Reviewer notes

- The distance heuristic lives in `AppKitDiffScrollerReconciler.scroll(to:)`, comparing the target offset against the current `scrollY` versus `viewportHeight`.
- It's opt-in via a new `AppKitDiffScrollRequest.snapsWhenFar` flag, set only for file-click requests (`AppKitDiffReviewScrollRequestResolver`). Inline-comment and draft-comment jumps are unaffected and keep animating, since those are already short, local moves.
- Added reconciler tests for both branches of the heuristic (long jump snaps, short jump still animates) and resolver assertions that `snapsWhenFar` is set correctly per command kind.

### DIFF
--- a/Alas/Sources/Center/Diff/Scroller/AppKitDiffRowSpec.swift
+++ b/Alas/Sources/Center/Diff/Scroller/AppKitDiffRowSpec.swift
@@ -70,6 +70,10 @@ struct AppKitDiffScrollRequest: Equatable {
     let alignment: AppKitDiffScrollAlignment
     let animated: Bool
     let generation: Int
+    /// When true, a long jump snaps instantly instead of animating even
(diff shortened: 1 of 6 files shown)


mrmans0n:
@codex review

PR #997, opened 2026-08-10
fix(acp): mention folder picker

mrmans0n:
## Summary

- prioritize exact, prefix, and contiguous relative-path matches before fuzzy results
- deduplicate mention candidates while preserving directory URLs
- make arrow keys, Enter, Tab, and Escape work from the focused search field
- stop hover selection from programmatically recentering the scroll view

## Root cause

Folder candidates could be duplicated with file-shaped URLs, direct path matches had no priority over scattered fuzzy matches, and key handling sat outside the focused text field. Hover selection also shared the same scroll-to-highlight path as keyboard navigation, causing pointer movement to fight manual scrolling.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/ACPMentionPickerTests -quiet`
- full test suite: 7,836 tests ran; three unrelated transient failures passed on isolated rerun

### DIFF
--- a/Alas/Sources/ACP/UI/ACPMentionPicker.swift
+++ b/Alas/Sources/ACP/UI/ACPMentionPicker.swift
@@ -217,7 +217,7 @@ enum MentionFuzzy {
         var result: [URL] = []
         var indices: [String: Int] = [:]
         for file in files {
-            let path = relativePath(for: file, root: root).lowercased()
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #1004, opened 2026-08-10
fix(review): open restored comment editors

mrmans0n:
## Summary

- redraw AppKit-hosted review rows when restored comment edit or reply mode changes
- include editor mode in row equality while keeping text-entry updates non-structural
- cover restored-window editing plus presentation-state and row-token invalidation

## Root cause

After restoring review state, the shared editor mode changed when Edit was pressed, but the hosted row's equality token stayed unchanged. The fresh row therefore skipped its redraw until an unrelated invalidation occurred.

## Validation

- post-rebase focused regression suites: 48 passed, 0 failed
- SwiftFormat lint and diff check
- macOS app build
- full test run: 7,932 passed, 12 skipped, 4 pre-existing unrelated failures

The unrelated full-suite failures were in ACP terminal orphan release, ACP auth launch quoting, feedback file filtering, and cleared base-branch override refresh coverage.


mrmans0n:
@codex review

PR #1005, opened 2026-08-10
fix(review): Highlight review replies

mrmans0n:
## Summary
- use the theme accent color for reply author labels
- use the same accent for reply rails

## Verification
- xcodegen
- macOS app build
- full test suite executed; four pre-existing unrelated tests failed


mrmans0n:
@codex review

PR #1011, opened 2026-08-11
feat(missions): gate hidden runtime

mrmans0n:
## Summary

- Hides Missions behind a debug-only feature flag that defaults off.
- Prevents hidden Mission UI from appearing in the sidebar, center tabs, and startup restoration paths.
- Suspends Mission runtime work while the feature is disabled, while preserving existing Mission state for users who re-enable it.

### Reviewer notes

- The feature flag lives in debug settings and propagates through `AppState` so UI and runtime behavior use the same source of truth.
- Mission-specific tests opt in explicitly where they need the legacy enabled behavior.

### Verification

- `rtk xcodebuild ... -only-testing:AlasTests/MissionsFeatureFlagTests -only-testing:AlasTests/MissionFeatureRuntimeTests`
- Included in the final Stack A focused suite and full-suite run before sync.

Part of stack `missions`

Commit: 60fbeb8


mrmans0n:
This change is part of the `missions` stack:

- 👉 #1011
- #1012
- #1013

PR #1015, opened 2026-08-11
fix(gg): recover stacks while detached

mrmans0n:
## Summary
- recover GG stack presentation after `gg mv` or `gg first` detaches HEAD
- invalidate stale detached stack state per affected worktree while preserving GG policy gates
- replace interrupted cold detached recovery for the active pane

## Testing
- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination platform=macOS -quiet build`
- focused `RightPaneGGStackTests`, `GGWorktreeContextTests`, and `AppStateGGACPWorktreeContextTests`

The full local suite was also exercised; it has existing order-dependent ACP/DiffReview failures outside this GG change, so CI is the final integration gate.


mrmans0n:
@codex review

PR #1016, opened 2026-08-12
perf(diff): cache pinned diff rows in appkit mode

mrmans0n:
## Summary

Cache pinned diff-row IDs when the render plan changes, so viewport updates do not rescan every row.

Adds a 20,000-row regression test that verifies scrolling does not repeat retention inspection.

## Validation

- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build`\n- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet -only-testing:AlasTests/AppKitDiffScrollerStressTests test`\n\nThe full suite was attempted but hangs in existing ACP test lifecycle infrastructure; the ACP classes pass individually.


mrmans0n:
@codex review

PR #1017, opened 2026-08-12
feat(worktrees): move attach issue to dialog header

mrmans0n:
## What

The `Attach issue...` button in the New worktree dialog sat between the Stacked Diffs Mode description and the "Open after create" field, where it read as part of neither section and was easy to miss.

It's now an icon-only paperclip button in the dialog header's top-right corner, with a tooltip and hover feedback.

## How

- `DialogContainer` gains an optional `headerAccessory` view builder, rendered top-right opposite the title/subtitle. Existing call sites are untouched via a `where HeaderAccessory == EmptyView` convenience init.
- New `DialogHeaderIconButton`: 26×26 icon-only button, glyph goes `fg-muted` → `fg` on hover with a `bg-3` fill and `line` border appearing, `.help()` tooltip, full-rect hit target.
- `NewWorktreeDialog` renders the button only while no issue is attached. Once an issue is attached, the existing issue card in the body keeps the Edit / Remove / Open actions.

## Testing

- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' -configuration Debug build` — succeeds.
- No test references the changed views beyond `DialogContainerLayout` constants, which are unchanged.

### DIFF
--- a/Alas/Sources/Dialogs/DialogChrome.swift
+++ b/Alas/Sources/Dialogs/DialogChrome.swift
@@ -5,10 +5,11 @@ enum DialogContainerLayout {
     static let projectWidth: CGFloat = 640
 }
 
-struct DialogContainer<Content: View>: View {
+struct DialogContainer<Content: View, HeaderAccessory: View>: View {
     let title: String
     let subtitle: String?
     let width: CGFloat
+    @ViewBuilder let headerAccessory: () -> HeaderAccessory
     @ViewBuilder let content: () -> Content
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

mrmans0n:
Addressed in 554d6313 — added `.accessibilityLabel(tooltip)` to `DialogHeaderIconButton` so VoiceOver announces the action rather than the symbol.

@codex review

PR #1019, opened 2026-08-13
fix(commit): keep draft editor focused while typing

mrmans0n:
## Summary

- keep the draft commit subject and description focused while their text bindings update
- cover repeated subject keystrokes with a regression test

## Root cause

The AppKit-backed composer controls were bridged through a manually managed `@FocusState`. Text updates caused that focus state to be reconciled as SwiftUI focus, which could resign the native field editor after one or two keystrokes. Ordinary `@State` keeps the visual focus selection stable while AppKit remains the source of truth for first-responder ownership.

## Validation

- manually verified in a freshly built Alas instance
- `xcodegen`
- focused repeated-keystroke regression test

The full test suite was also attempted, but Xcode's test runner stalled while finalizing the test session and was interrupted after 699 seconds without reporting a test failure.

### DIFF
--- a/Alas/Sources/Center/Commit/CommitMessageEditorView.swift
+++ b/Alas/Sources/Center/Commit/CommitMessageEditorView.swift
@@ -17,7 +17,7 @@ struct CommitMessageEditorView: View {
     var accessory: AnyView? = nil
 
     @Environment(\.theme) private var theme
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #1020, opened 2026-08-13
fix(gg): stack query timeout

mrmans0n:
## What changed

Use the GG runner's existing 600-second command timeout for `gg ls --json`, rather than the generic 30-second process timeout.

## Why

On larger stacks, `gg ls --json` can take longer than 30 seconds. The timeout cleared the GG stack state and made Alas fall back to the ordinary commit list, hiding stack commits above the checked-out position.

## Validation

- `xcodegen`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet test -only-testing:AlasTests/GGServiceTests`

The complete local suite was interrupted after two unrelated, reproducible `AppStateCLIRoutingTests` failures (`cliWorktreeNewUsesDialogPreferredBaseWhenBaseIsOmitted` and `cliWorktreeDeleteMarksMatchedWorktreeDeleting`).

### DIFF
--- a/Alas/Sources/Integrations/GG/GGService.swift
+++ b/Alas/Sources/Integrations/GG/GGService.swift
@@ -63,24 +63,17 @@ struct ProcessGGCommandRunner: GGCommandRunning {
     }
 
     func run(args: [String], cwd: URL?) async throws -> ProcessResult {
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #1022, opened 2026-08-13
fix(acp): chat scroll jumps after reading

mrmans0n:
## Summary

- pause transcript tail-follow as soon as a verified upward user scroll leaves the bottom tolerance
- keep layout-driven bounds changes from disabling tail-follow
- add a delayed-settle regression test for small upward scrolls

## Verification

- `xcodegen`
- focused ACP scroll classifier, live-scroll, and scroll-back test suites (35 tests)
- `swiftformat --lint` on the five changed files
- `git diff --check`

## Manual testing

- launched the local Debug build and confirmed the non-streaming ACP chat stays where the reader stops after a small upward scroll


mrmans0n:
@codex review

PR #1031, opened 2026-08-14
feat(treesitter): Repack tree-sitter grammars into a Rust staticlib

mrmans0n:
## Summary

Replaces 22 SwiftPM packages and six vendored grammar trees (`ThirdParty/TreeSitter*`) with `ThirdParty/treesitter-pack`, a single Rust crate that depends on the grammars via cargo and exposes languages and highlight queries over a small C ABI (`alas_ts_language`, `alas_ts_query`). Queries are compiled into the binary, so `LanguageRegistry` no longer scans SPM resource bundles at runtime.

The six vendored trees existed only to route around a bug in each grammar's `Package.swift` (a relative-path `FileManager` check that breaks under SPM resolution, or missing query resources), not around a problem with the grammar itself — the equivalent `bindings/rust` in every one of those repos works correctly. This also drops ~9 MB of generated C from the repo and six near-identical `update-treesitter-*.sh` re-vendoring scripts.

Kotlin moves off the stale `fwcd/tree-sitter-kotlin` grammar (last release Aug 2024) to the maintained `tree-sitter-kotlin-ng`. That grammar ships no highlight query at all — confirmed via a recursive search of its full history — so `queries/kotlin/highlights.scm` is hand-written against its `node-types.json`/`grammar.js`, verified with a real `Query` compilation against the `TSLanguage` plus a runtime smoke test that parses real Kotlin source and asserts the expected capture categories actually fire.

## Verification

- All 25 grammar entry points confirmed *defined* in the linked archive via `nm`, alongside both FFI exports, with zero leftover SPM grammar resource bundles.
- Full app build succeeds; all 44 `TreeSitterHighlighter` tests pass unchanged, including the existing `kotlinBasics()` test against the new grammar.
- Crate test suite (6 tests): every language resolves, every query is non-empty and compiles, plus the two Kotlin-specific tests above.
- `build-treesitter-pack.sh`'s own test suite passes (fast path, fingerprint invalidation, missing-symbol guard) and its own fingerprint hashes the crate's own files, so an uncommitted edit invalidates the cached artifact.
- `SwiftTreeSitter` stays as the only tree-sitter SwiftPM dependency — it's the runtime and Swift API, not a grammar, so there's no symbol collision.

Design doc: `docs/superpowers/specs/2026-08-14-treesitter-pack-design.md` (gitignored, not part of this diff).


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #1045, opened 2026-08-19
feat(diff): Improve diff scrolling and highlighting

mrmans0n:
## Summary

- Batch diff row height corrections and avoid redundant frame assignments and SwiftUI layouts while scrolling.
- Preserve the first real layout for new or reused rows so diff contents never appear as empty black boxes.
- Apply line-level tree-sitter fallback tokens to changed lines when parsing a partial existing-file hunk misses syntax captures.
- Add regression coverage for geometry batching, stable scrolling, initial rendering, and changed-line highlighting.

## Root cause

Visible diff rows were individually retiling and reassigning unchanged host frames during scroll, causing every mounted SwiftUI host to lay out on each scroll event. Separately, tree-sitter parses of partial hunks can produce incomplete captures for changed lines in existing files.

## Verification

- `xcodegen`
- Focused `AppKitDiffScrollerReconcilerTests`, `AppKitDiffScrollerStressTests`, and `DiffPaneViewTests`
- Manual side-by-side validation of initial rendering and scrolling

A full local suite run was interrupted after failures outside this diff in remote-web cache-version and GG stack tests, followed by an Xcode test-runner hang.

### DIFF
--- a/Alas/Sources/Center/Diff/DiffPaneTextDocumentBuilder.swift
+++ b/Alas/Sources/Center/Diff/DiffPaneTextDocumentBuilder.swift
@@ -725,6 +725,7 @@ struct DiffPaneTextDocumentBuilder {
                 lines: document.lines,
                 startIndex: segment.startIndex,
                 endIndex: segment.endIndex,
+                fileExtension: fileExtension,
                 theme: theme
             )
         }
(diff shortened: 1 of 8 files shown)


mrmans0n:
@codex review

PR #1047, opened 2026-08-20
fix(gg): Show GG amend failures in Changes

mrmans0n:
## Summary

Makes GG safety failures visible in the Changes tab, so a failed staged amend no longer looks like it silently did nothing.

## Changes

- Pass the existing GG mutation error into the preparation model.
- Show it below the GG actions and keep that card visible.
- Add regression coverage for the error state.

## Testing

- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination "platform=macOS" -quiet build`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination "platform=macOS" -quiet test -only-testing:AlasTests/ChangesPreparationModelTests`

## Notes for reviewers

A full suite run exposed unrelated existing failures and then hung during Xcode test cleanup.

### DIFF
--- a/Alas/Sources/Right/ChangesPreparationCard.swift
+++ b/Alas/Sources/Right/ChangesPreparationCard.swift
@@ -53,6 +53,13 @@ struct ChangesPreparationCard: View {
                         ggDestinationButton(action)
                     }
                 }
+                if let error = model.mutationError {
+                    Text(error)
+                        .font(.system(size: 11))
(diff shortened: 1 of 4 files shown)


mrmans0n:
Codex review: 👍 No findings. The mutation error now reaches the visible Changes preparation state, and the regression test covers the previously hidden failure path.

PR #1050, opened 2026-08-21
fix(changes): stale changes after GG mutations

mrmans0n:
## Summary

Refresh Git changes before waiting for the GG stack after a mutation, so a successful amend immediately clears its staged rows instead of leaving the UI stale while GG reloads.

## Changes

- Refresh Git status before the GG stack for non-clean GG mutations.
- Add a regression test that holds the stack refresh and verifies the changes refresh occurs first.

## Where to start

- `Alas/Sources/Integrations/GG/GGMutationCoordinator.swift` — shared post-mutation refresh ordering.

### DIFF
--- a/Alas/Sources/Integrations/GG/GGMutationCoordinator.swift
+++ b/Alas/Sources/Integrations/GG/GGMutationCoordinator.swift
@@ -516,8 +516,8 @@ final class GGMutationCoordinator {
             return
         }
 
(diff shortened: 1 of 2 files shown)


mrmans0n:
Codex review: 👍 No findings. Git changes now refresh before a slow GG stack reload, and the regression test proves staged changes update while that reload is pending.

PR #1051, opened 2026-08-21
feat(diff): Speed up stack entry picker

mrmans0n:
Summary

- Reuse the already-loaded Changes gg stack when opening the follow stack entry picker, so the sheet can populate immediately when the data is fresh.
- Render PR numbers with verbatim Text so SwiftUI does not format them.
- Add a focused AppState regression test for the fast path.

Testing

- xcodegen
- xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' test -only-testing:AlasTests/AppStateFollowStackEntryTests -quiet
- Full local xcodebuild test hit existing RightPaneGGStackTests failures and then Xcode hung during test cleanup.


mrmans0n:
👍 No findings from my review. I checked the fast-path guard, the verbatim PR number rendering, and the regression test against the branch diff. CI is green.

PR #1060, opened 2026-08-24
fix(chat): input-request scroll jumps

mrmans0n:
## Summary

Keep an ACP transcript’s temporary history window stable while an input form is pending, so the delayed scroll-settle callback cannot move the message being read.

## Changes

- Skip window compaction while user input is pending.
- Add regression coverage for the delayed post-scroll callback.

## Verification

- `xcodegen`
- `swiftc -parse AlasTests/ACP/UI/ACPTranscriptScrollerLiveScrollTests.swift Alas/Sources/ACP/UI/Scroller/ACPTranscriptScroller.swift`
- `git diff --check`

The full macOS Xcode test run is currently blocked in this worktree before compilation because Xcode rejects the generated `.build/ghostty/GhosttyKit.xcframework`.


mrmans0n:
@codex review

PR #1066, opened 2026-08-25
fix(worktree): branch name field focus

mrmans0n:
## Summary
Restores editability on the custom AppKit text-field cell so branch names can receive focus when creating a worktree. Adds a regression test for first-responder editing.

## Changes
- Restore editable and selectable state after installing the custom cell.
- Verify the field becomes first responder and creates its editor.


mrmans0n:
@codex review

PR #1069, opened 2026-08-25
fix(diff): fallback glyph diff row heights

mrmans0n:
## Summary

Split diff rows could repeatedly add height when an AppKit fallback glyph made the measured line taller than its paragraph line height. Correct existing padding from the measured delta so the layout converges.

## Verification

- Reproduced with the Android ScoutOrchestrator progress glyph.
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -only-testing:AlasTests/DiffPaneViewTests/splitRowHeightSyncConvergesForUnicodeFallbackGlyph\(\) test`\n- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build`\n\n## Where to start\n\n- `Alas/Sources/Center/Diff/DiffPaneTextDocumentView.swift` — split-row height correction.\n- `AlasTests/DiffPaneViewTests.swift` — exact fallback-glyph regression.

### DIFF
--- a/Alas/Sources/Center/Diff/DiffPaneTextDocumentView.swift
+++ b/Alas/Sources/Center/Diff/DiffPaneTextDocumentView.swift
@@ -871,9 +871,16 @@ final class DiffPaneTextScrollView: NSScrollView {
             var candidateLineHeight: CGFloat?
             var lineCount = 1
             if let targetHeight {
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #1117, opened 2026-09-04
fix(acp): hide provider pill when there's nothing to switch to

mrmans0n:
## Summary

- The composer's provider pill (`Provider: main`) is purely informational, so it should only take up space when the adapter actually offers a choice between providers.
- Extends the existing "hide the lone required provider" rule (#1068) to hide the pill whenever there's at most one *active* provider — required or not.
- Ignores disabled providers (no `current` config) sitting alongside the active one when deciding whether a real choice exists, so a single usable provider with a disabled sibling still hides the pill.

### Reviewer notes

- `ACPSession.currentProviderDisplayName` now filters to providers with `current != nil` first, then requires more than one before showing anything — same shape as before, just gated on count rather than the required-flag special case.
- Added coverage in `ACPSessionAgentStateTests` for the lone-optional-provider and disabled-sibling cases.

### DIFF
--- a/Alas/Sources/ACP/Session/ACPSession.swift
+++ b/Alas/Sources/ACP/Session/ACPSession.swift
@@ -197,12 +197,15 @@ final class ACPSession: ObservableObject, Identifiable {
         return availableModels.first { $0.id == id }?.name ?? id
     }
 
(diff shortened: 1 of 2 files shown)


mrmans0n:
@codex review

PR #1118, opened 2026-09-05
perf(acp): skip unchanged transcript scroll layout passes

mrmans0n:
Scrolling the ACP transcript revisited every mounted row on each tick, even when the mounted range was unchanged. Taller viewports increased these redundant content comparisons and frame assignments.

Skip those layout passes until the mounted range changes. Content updates, resizing, and intrinsic height changes still run full layout, and the cached range is published only after reentrant layout converges.

Validation:
- In 20 small scroll steps, row comparisons fell from 560 to zero at an 800-point viewport and from 800 to zero at 2160 points.
- macOS build and SwiftFormat lint passed.
- 96 tests across 10 scroller suites passed, including band crossings, retained offscreen forms, content growth, and mount-time remeasurement.
- The full local suite stalled with terminal-cleanup workers blocked after three worktree-creation timeouts. Those three tests and the stalled Mermaid test all passed in a fresh, isolated run.

Actual 4K frame-rate improvement has not been profiled yet.


mrmans0n:
@codex review

PR #1018, opened 2026-08-12
feat(diff): Productionize AppKit diff scrollers

mrmans0n:
## Summary

- make AppKit diff scrolling the unconditional production path for standalone diffs, reviews, and GG split previews
- remove the rollout flag and Advanced settings control while retaining unreachable legacy implementations for later deletion
- preserve the static-height SwiftUI path and add regression coverage for the production route

## Validation

- `xcodegen`
- focused AppKit diff-scroller and stress test suite
- Debug build
- Release build

The full test suite has pre-existing worktree-operation timeouts when run together; the failing CLI operation test passes independently on both the base and feature branches.

PR #1119, opened 2026-09-05
feat(appkit): Productionize the AppKit diff scroller and delete the legacy SwiftUI paths

mrmans0n:
## Summary

- Removes `AppKitChangesScrollerFlag`, making the virtualized AppKit scroller the only implementation for the Changes tab (it's been the debug default since #1044).
- Deletes the SwiftUI diff/review scrolling code that went unreachable when #1018 removed `AppKitDiffScrollerFlag` but left the old implementations in the tree: `DiffPaneView`'s legacy scroll body, `DiffReviewSurface`'s legacy review stream, and the `DiffReviewFileSection` view (~940 lines) plus its now-dead collaborators.
- Renames the remainder of `DiffReviewFileSection.swift` to `DiffReviewComments.swift` (draft/thread comment cards, composers, placement logic — shared by the AppKit row plan) and `StashesSectionView.swift` to `StashRows.swift`.

### What changed

- **Changes tab**: `ChangesTabView` always renders `AppKitDiffScroller`; the SwiftUI `scrollContent` fallback, the `StashesSectionView` composition it alone used, and the Settings → Advanced toggle are gone.
- **Diff pane**: `legacyInternalScrollBody` and `lazyRowsStack` (a byte-identical twin of the still-live `staticRowsStack`) removed.
- **Review surface**: `legacyMainReviewStream`, its exclusive scroll helpers, and the now-orphaned SwiftUI visibility-sentinel members of `DiffReviewSurfaceSelectionSync` removed.
- **`DiffReviewFileSection`**: deleted along with its `.equatable()` wrapper and every type whose only caller was it or a `DiffReviewSurface.fileSection(_:automaticallyRendersDiff:)` helper that had itself gone unreachable in the same prior cleanup. `ReviewChangesFileSection`'s dead typealias and `DiffPaneLSPContext.rendersEqual` (superseded by the AppKit token decomposing LSP context into comparable fields) go too.

### Test coverage

Ported test cases with an AppKit-side equivalent onto the production path (`AppKitDiffReviewScroller` / `AppKitDiffReviewRowInput`), matching how `DiffReviewSurface` and `ReviewChangesTabView` actually render. New row-equality-token tests in `AppKitDiffReviewRowPlanTests` cover display preferences, action presence, and draft/inline-feedback availability changing independently of the comment/feedback itself — the AppKit equivalent of what `DiffReviewFileSectionEqualityTests` (deleted) checked for the legacy view.

### Reviewer notes

Two legacy assertions didn't survive the port; called out inline rather than dropped silently, both flagged as follow-up investigations rather than blockers here:
- A materialization test scrolling a single large accessory hunk (one hunk, many per-line draft comments) to its last comment never realizes the target row under the AppKit scroller — reproduced via both the scroll-command path and a direct `setScrollY`, regardless of drain time. Looks like a real virtualization gap for that specific row shape, not a test-porting mismatch.
- The "Copied prompt" confirmation chip: `DiffReviewSurface` wraps `copyPrompt` with `state.copyFeedback.show(...)`, but nothing forwards `CopyFeedbackState`'s own `@Published` change into `AppKitDiffReviewPresentationStore`'s `structuralDidChange`, so the chip never renders via the AppKit path.

Also worth noting for anyone porting similar tests in the future: interaction tests need to mount through `DiffReviewSurface` (or bridge `AppKitDiffReviewFileState.structuralDidChange` into a `@StateObject`) rather than a bare `AppKitDiffReviewScroller` — a state-only change with no row-structure change won't otherwise trigger a rebuild.


mrmans0n:
@codex review

PR #1121, opened 2026-09-05
feat(editor): code boxes for triple-backtick fences in markdown editors

mrmans0n:
## What

Typing three backticks in a markdown-producing editable surface now expands into a fenced code box — a tinted, bordered, monospaced block that matches how the same fence renders once sent, instead of leaving four stray backticks behind.

Covers:
- The ACP chat composer (with Enter staying inside the box instead of submitting, and ⌘⏎ added as an always-works send shortcut)
- The commit message body editor
- Review comment editors (inline diff comments, reply drafts, the review summary rail, the verdict sheet, the publish confirmation dialog)
- The Settings prompt editor (Commit / Merge-Resolve AI prompts)

Deliberately out of scope: syntax highlighting inside the box, hiding/replacing the fence markers, a language picker, and any change to how the transcript itself renders fenced blocks.

## How

- `MarkdownFenceEditing` (new) parses fenced blocks from plain text and decides what a given keystroke should do to them.
- `MarkdownCodeBlockStyler` (new) applies the box's text attributes; `PairedDelimiterTextView` (existing AppKit base class shared by the composer and every other editor here) draws the box itself via `drawBackground(in:)`, since `.backgroundColor` paints behind glyphs only.
- The trickiest part — deciding whether a keystroke should be swallowed to prevent it from re-cutting an existing fence's structure — ended up going through a few iterations before landing on the right shape: instead of hand-checking backtick counts and line positions per case, it now simulates the exact edit AppKit's own pairing logic would make and asks the real parser whether the result would change the document's fence structure. That's a small function, but it's carrying the weight of the whole feature's safety guarantee, so it got adversarial review from a few different angles (selection edges, dead-key keyboard layouts, non-backtick auto-paired characters) before settling.

## Testing

Local focused suites are green (180+ tests across the touched areas). Full sweep left to CI per this repo's convention — this machine is slow for `xcodebuild test` unfiltered.

One pre-existing, unrelated failure exists on `main` and is unaffected by this branch: `CommitMessageEditorViewTests.pairedSubjectAndBodyWrapSelectedText()`. Verified independently two ways — reverting this branch's changes in place reproduces the identical failure, and the diff to both files it touches is purely additive (no existing line changed), so it couldn't be the cause.

## Known follow-up (not in this PR)

`ReviewDraftComposerNSTextView` — the actual "leave a comment on this diff line" composer used in `DiffTabView.swift` / `DiffReviewFileSection.swift` / `AppKitDiffReviewRowPlan.swift` — is a bespoke subclass of the same base class the rest of this feature builds on, and it was missed from this PR's scope (the surface inventory was assembled by grepping for the convenience wrapper this feature hooks into, which this class doesn't use). Its sibling reply editor gets the feature; this one doesn't yet. Tracking as a follow-up rather than expanding this PR.

### DIFF
--- a/Alas.xcodeproj/project.pbxproj
+++ b/Alas.xcodeproj/project.pbxproj
@@ -957,6 +957,7 @@
 		9B00878F781FADB428684B2D /* DiffDisplaySignatureTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 283613E170E25F69521F64ED /* DiffDisplaySignatureTests.swift */; };
 		9B6940D68B098D1F1305324B /* GitTypesTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 271F79A165CAA8617C1EAA4E /* GitTypesTests.swift */; };
 		9BA16B0D5FC26501D54B5192 /* WorkingTreeStageAllActionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F931B1EEA7F04855BAE5A43 /* WorkingTreeStageAllActionTests.swift */; };
+		9BAF14A88AEB1F888D7541DC /* MarkdownCodeBoxSurfaceOptInTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6ED105A2C41B953A4DC9B8F6 /* MarkdownCodeBoxSurfaceOptInTests.swift */; };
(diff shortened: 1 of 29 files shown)


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #1126, opened 2026-09-06
fix(review): cap outdated drawer height so long lists scroll instead of overflowing

mrmans0n:
## Problem
In the PR Review pane, expanding the "Outdated & file-level" drawer with many threads made the list render at its full, uncapped content height and overflow past the drawer into the rest of the review pane — with nothing left to scroll.

## Root cause
`.fixedSize(vertical: true)` forced the list's `ScrollView` to size to its ideal (full content) height, which the surrounding `.frame(maxHeight:)` only clamped for layout accounting purposes, not actual rendering — so long lists ignored the cap entirely.

## Fix
Measure the list's actual content height with `onGeometryChange` and set the `ScrollView`'s frame explicitly to `min(measuredHeight, cap)`, with `.clipped()` as a backstop. Short lists still shrink-wrap; long lists stay capped and scroll properly.

## Testing
- Added a regression test reproducing the overflow (measured 1892pt before the fix vs. the intended 280pt cap).
- Added pure-function coverage for the capping arithmetic.
- `xcodebuild test` passes for the affected suites; full `xcodebuild build` succeeds.

### DIFF
--- a/Alas/Sources/Center/Review/OutdatedThreadsDrawer.swift
+++ b/Alas/Sources/Center/Review/OutdatedThreadsDrawer.swift
@@ -10,12 +10,20 @@ enum OutdatedThreadsDrawerPresentation {
 
         return min(defaultMaxExpandedListHeight, max(80, floor(availableHeight * 0.35)))
     }
+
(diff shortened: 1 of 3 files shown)


mrmans0n:
@codex review

PR #1145, opened 2026-09-09
fix: retain landing cancellation before task attachment

mrmans0n:
Part of stack `gg-land-ui`

Commit: 648059b


mrmans0n:

PR #1147, opened 2026-09-09
feat: show live landing in a tab

mrmans0n:
Part of stack `gg-land-ui`

Commit: 044f199


mrmans0n:

PR #1149, opened 2026-09-09
fix: keep landing status across checkouts

mrmans0n:
- Resolve active repository landing before GG drawer policy
- Cover inactive GG context during an active landing


mrmans0n:

PR #1151, opened 2026-09-09
fix: retain landing scope and drain cancellation

mrmans0n:
- Preserve the confirmed scope across repeated restart attempts.
- Dispatch interruption through the runner protocol and retain final events.
- Repair callback and fixture compilation and cover restart regressions.


mrmans0n:

PR #1166, opened 2026-09-10
fix(acp): preserve queue when steering

mrmans0n:
## Summary

Composer steering now cancels the active turn without deleting pending queued prompts. The explicit queued-item “Send Now” action retains its existing queue-replacement behavior.

## Changes

- Preserve pending queue entries during normal `.steer` submission.
- Remove only a cancelled `.sending` queue head.
- Keep destructive queue replacement and undo for queued-item Send Now.
- Add regression coverage for both paths and update steering copy.

## Where to start

- `Alas/Sources/ACP/Session/ACPSessionRunner.swift` — routing behavior
- `AlasTests/ACP/Session/ACPSessionRunnerQueueTests.swift` — regression cases

## Verification

- `xcodegen`
- `xcodebuild … build`
- Focused ACP queue suite: 29 tests passed


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex Please add an explicit thumbs-up if the current commit is approved.

PR #1168, opened 2026-09-10
fix(mcp): improve MCP registration retries

mrmans0n:
## Summary

MCP hello attempts were exhausted well before the ACP registration grace period, leaving the warning affordance visible even when the server became available shortly afterward.

- Retry at roughly 0, 4, and 8 seconds, within the existing 12-second grace period.
- Cover a delayed socket registration that would have failed under the old schedule.
- Present reconnect as an outlined button with a refresh icon.

## Validation

- `cargo test -p alas-client`
- `xcodebuild -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS' -quiet build`\n- Full app suite ran 8,954 tests; 23 unrelated pre-existing failures remain.


mrmans0n:
@codex review

PR #1189, opened 2026-09-12
style(composer): unify the Send split button and stop it shrinking

mrmans0n:
## What

Finishes the segmented-button work started in 6e41b61, which only reached `queueSplitCapsule` — the Send button, the one that actually mattered, was untouched.

Target look is Slack's segmented send control: one continuous fill across both halves, split by a thin vertical divider before the chevron.

### Changes

- **`sendCapsule`**: `spacing: 1` → `0`, plus the inset hairline divider on the trailing edge of the primary half. It can't reuse `theme.color("line")` the way Queue does — that's a neutral hairline meant for neutral fills and it disappears against `accent` — so Send tints its own foreground (`bg-0`) instead.
- **Chevron fill moved from the Menu's label to the Menu itself.** With `.menuStyle(.borderlessButton)` the label is wrapped in the control's own chrome, so a background on the label stops short of the edges and leaves a visible gap next to the primary half. This is why the chevron rendered detached on the dark composer background rather than on the accent fill.
- **Shared `segmentDivider` helper + divider metrics** in `ACPComposerActionButtonMetrics`. The two capsules diverged in the first place because the divider was an inline literal in only one of them.
- **Action button no longer shrinks.** `fixedSize(horizontal:)` + `layoutPriority(1)` in the composer toolbar: Send/Queue/Stop is the row's primary action and shouldn't be what ellipsizes when the composer is narrow. The parameter chips truncate instead.

## Testing

- `xcodebuild build` — clean.
- `ACPComposerActionButtonMetricsTests` — 3/3 pass, including a new assertion that the divider stays a sub-capsule-height hairline at partial opacity.

The visual result has not been verified on-device yet; these are layout changes whose failure mode is precisely "reads fine, renders wrong", so the rendering deserves an eyeball.

Note: `AlasTests/RemoteWebAssetTests.swift:565` fails on this branch independently of these changes (force-unwrap of a missing `/changes-view.js?v=6` range returns nil and crashes the test host). No remote/web files are touched here.

### DIFF
--- a/Alas/Sources/ACP/UI/ACPComposerActionButton.swift
+++ b/Alas/Sources/ACP/UI/ACPComposerActionButton.swift
@@ -86,6 +86,10 @@ struct ACPComposerActionButton: View {
                     .font(.system(size: 9, weight: .semibold))
                     .foregroundStyle(theme.color("bg-0"))
                     .padding(.horizontal, 7)
+                    // Height stays on the label so the whole painted segment is
+                    // clickable — a frame applied after `Menu` grows the layout
+                    // and background but leaves the hit target at label height.
+                    .frame(height: ACPComposerActionButtonMetrics.capsuleHeight)
             }
             .menuStyle(.borderlessButton)
             .menuIndicator(.hidden)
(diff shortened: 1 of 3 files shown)


mrmans0n:
Good catch on the chevron hit target — that was a real regression in this PR, not a pre-existing quirk. Applying `.frame(height:)` after `Menu` grew the layout and the painted fill while the control kept its label-height hit target, so the top and bottom of the segment would have been dead pixels that look clickable.

Moved the height back onto the label in both the Send and Queue menus (77adfd1b). The fill stays on the outside of the `Menu`, since that part was deliberate: with `.borderlessButton` the label sits inside the control's own chrome, so a background on the label stops short of the edges and leaves the gap this PR is trying to close.

@codex review

PR #1191, opened 2026-09-12
feat(sidebar): show delegated sessions as a tree in the agent tab

mrmans0n:
## Summary
- Delegated child sessions (via `alas session new`) rendered as flat, indistinguishable cards in the agent sidebar, which looked like session duplication when several delegations landed close together with fast-changing auto-generated titles.
- Nests a delegated child directly under its parent when both land in the same sidebar section, with a connector rail; a parent card shows a "N delegated" count pill.
- When a family straddles Active/History (parent idle, child still running, or vice versa), the child stays at top level with a "Delegated by <parent>" caption instead of being hidden or misplaced.
- Warms `delegatedSessionParents` from the orchestration store at app startup, so hierarchy survives a restart instead of only covering delegations recorded in the current session.

### Reviewer notes
- `AgentSidebarRollupBuilder.nestDelegations` assumes delegation is at most two levels deep — `ACPSessionOrchestrationPolicy.authorizeCreate` already refuses to let a session with a parent create a child, so this is enforced elsewhere, not re-validated here. A defensive check still guards against a corrupt parent cycle by falling back to a flat caption rather than dropping rows.
- New `ACPOrchestrationStore.delegationParents()` / `ACPOrchestrationPersistence.delegationParents()` return every child→parent link regardless of phase, used only for this startup warm-up.
- Terminal rows are explicitly excluded from delegation annotation (delegation only applies to ACP sessions).

### Testing
- `AgentSidebarRollupTests`: nesting within a section, cross-section fallback caption, orphaned-parent fallback, live row replacing a persisted row nests exactly once, mutual-reference cycle stays flat, terminal rows never annotated.
- `ACPOrchestrationStoreTests`: `delegationParents()` returns links across all phases and is empty for a fresh store.

### DIFF
--- a/Alas/Sources/ACP/Orchestration/ACPOrchestrationPersistence.swift
+++ b/Alas/Sources/ACP/Orchestration/ACPOrchestrationPersistence.swift
@@ -41,6 +41,10 @@ actor ACPOrchestrationPersistence {
         try openedStore().parent(childSessionId: childSessionId)
     }
 
+    func delegationParents() throws -> [String: String] {
+        try openedStore().delegationParents()
+    }
+
     func updatePhase(
         childSessionId: String,
         phase: ACPDelegationPhase,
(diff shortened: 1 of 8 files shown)


mrmans0n:
@codex review

PR #1196, opened 2026-09-12
feat(attention): dismiss items from inbox and present it as a popover

mrmans0n:
## Summary

- Adds per-item dismissal to the needs-attention inbox: each active row gets a `Dismiss` button and the header gets a `Dismiss all` action when items exist. Dismissing acknowledges the underlying event, so the item moves to the *Earlier* section until its condition changes (same semantics as navigating to an item).
- Replaces the full center-pane inbox with a popover anchored to the sidebar toolbar button (~400pt wide, capped height, scrollable). The center pane no longer swaps content while the inbox is open.
- Removes the return-destination machinery (`AttentionReturnDestination`, save/restore on open/close, validation helpers) that only existed to restore center-pane state after closing the full-pane inbox.

## Reviewer notes

- `isAttentionInboxOpen` is kept with its `didSet` side effects (navigation generation bump, pending review reveal clear) — popover dismissal now routes through the SwiftUI binding.
- The three tests covering return-destination restore behavior were deleted along with the feature; a new test covers the simplified open/close toggle.
- Existing failures in `AppStateAttentionTests` (checkout-owned session flows, fingerprint encoding) are pre-existing on `main` and unrelated to this change.

### DIFF
--- a/Alas/Sources/App/AppState.swift
+++ b/Alas/Sources/App/AppState.swift
@@ -103,7 +103,6 @@ final class AppState {
     @ObservationIgnored var attentionNavigationDepth = 0
     var attentionNavigationErrors: [UUID: String] = [:]
     @ObservationIgnored var attentionNavigationEnvironment: AttentionNavigationEnvironment?
-    @ObservationIgnored var attentionReturnDestination: AttentionReturnDestination?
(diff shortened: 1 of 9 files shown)


mrmans0n:
@codex review

PR #1205, opened 2026-09-12
fix(minimap): layout crash in transcript and editor

mrmans0n:
## Summary

Fix a minimap layout loop that crashes the transcript while AppKit updates hosted SwiftUI rows. Each unchanged scroll-view tile pass previously expanded and then shrank the clip view, repeatedly invalidating layout.

Place the minimap beside the native scroll view in a shared container used by the transcript and editor. Preserve wheel forwarding, document navigation, and responsive visibility based on the full pane width.

## Validation

- App build and SwiftFormat lint passed.
- All 16 minimap tests passed before rebasing; rerunning on the rebased commit.
- Direct AppKit checks with hosted SwiftUI content and an NSWindow confirmed zero clip-width changes across repeated layout passes for legacy and overlay scrollers; the original implementation produced 20 changes across 10 passes.
- Regression coverage includes stable layout, reserved space, responsive visibility, wheel forwarding, rendering, and navigation.
- The local full suite reported failures in attention, worktree, and session tests outside the minimap suite, then stalled and was interrupted. CI results are pending.

### DIFF
--- a/Alas/Sources/ACP/UI/Scroller/ACPTranscriptScroller.swift
+++ b/Alas/Sources/ACP/UI/Scroller/ACPTranscriptScroller.swift
@@ -57,13 +57,13 @@ struct ACPTranscriptScroller: NSViewRepresentable {
 
     func makeCoordinator() -> Coordinator { Coordinator() }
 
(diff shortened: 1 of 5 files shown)


mrmans0n:
@codex review

PR #1212, opened 2026-09-13
fix(mcp): style Switch to HTTP button like Reconnect

mrmans0n:
## What
The "Switch to HTTP transport" button in the MCP status popover used `AlasButton(style: .subtle)`, which renders with no background and no border — so it read as plain text rather than an actionable button, unlike the "Reconnect session" button (`style: .normal`) right below it.

## Fix
Switch it to `style: .normal` to match Reconnect, and add a matching icon (`arrow.triangle.2.circlepath`) for visual parity.

## Testing
- `xcodebuild ... build` — succeeds
- `xcodebuild ... -only-testing:AlasTests/ACPMCPStatusPolicyTests test` — 11/11 pass


mrmans0n:
@codex review

PR #1214, opened 2026-09-13
fix(acp): make queue Send now non-destructive, redesign Up next rows

mrmans0n:
## Summary

- Fixes "queue disappears" when force-sending a queued item while the agent is busy: Send now now interrupts the running turn with the selected item and leaves every other pending item untouched, instead of discarding the rest of the queue behind a 5s undo toast.
- Adds a non-interrupting "Move to front" action for reordering a queued item to the head without touching a running turn.
- Redesigns the queued-message list: full-width "Up next" rows replace the right-aligned dashed bubbles, numbered by dispatch order with a "Next" tag on the head item, a hover toolbar (Move to front, Send now, Retry, Edit, Remove), and a right-click menu adding Move up/Move down so reordering isn't drag-only. The header now shows from a single queued item onward.
- Removes the discard-and-undo plumbing end to end (session, runner, manager, remote protocol, gateway, web client) since nothing is discarded anymore.

### Reviewer notes

- `ACPSessionRunner.steer(...)` no longer takes a `discardQueue` parameter; the only queue mutation on steer is dropping a `.sending` head, same as before.
- `ACPTranscriptQueuePolicy.canMoveQueueItem` is a new pure predicate shared by `ACPSession.moveInQueue` and the row's Move up/down enablement, so the two can't disagree about what's a legal reorder.
- Remote web keeps its existing UI; only the server-side semantics of `queueForceSend` changed (non-destructive) and the `steerUndoAvailable` wire field / toast were dropped.
- Local focused test suites across session, runner, policy, remote protocol/gateway, and scroller layers pass (354 tests). Full project build/test is left to CI per this repo's local machine constraints.


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #1220, opened 2026-09-13
feat(layout): animate sidebar and right-pane collapse

mrmans0n:
## Summary
- Adds motion to collapsing/expanding the left sidebar and the right pane, which previously snapped instantly.
- The pane's frame slides toward its window edge while its content lags behind and fades, so the divider visibly overtakes the content instead of wiping it away.
- One shared curve (`PaneCollapseMotion`) drives every trigger — menu shortcuts, toolbar buttons, right-rail clicks, and narrow-window auto-collapse — so they all move the same way. Reduce Motion disables the animation entirely.

### What changed
- `Layout/PaneCollapseMotion.swift` (new) — the shared spring curve plus the content-lag/fade transition; reuses the existing pane-drawer curve so the whole window shares one feel.
- `Layout/ThreePaneLayout.swift` — sidebar and the rail-disabled right pane animate on effective visibility, keyed from one place.
- `Right/RightPaneView.swift` — the rail-enabled right body slides under the fixed rail using the same motion, preserving the pane's existing identity guarantee.
- `Center/TabBarView.swift` — the "show sidebar" / "show right sidebar" toolbar buttons fade in instead of popping.
- `AlasTests/PaneCollapseMotionTests.swift` (new) — covers the offset math and the Reduce Motion path.

### Reviewer notes
- Gutter drags are intentionally untouched (they mutate width, not visibility).
- Worth a manual look at the right pane while an AppKit-backed view (terminal, diff scroller) is active, since those draw above SwiftUI content.


mrmans0n:
@codex review

PR #1233, opened 2026-09-15
perf(editor): reduce editor repaint and scroll-anchor layout work

mrmans0n:
## Summary

Syntax and diagnostic color updates currently rebuild the entire editor display, replace inlay-hint attachments, and clear temporary semantic colors. Apply paint-only changes to the affected source-backed display ranges instead, preserving attachments, selection, and overlays. Font, paragraph, and composition changes retain the existing rebuild path.

Scroll-anchor restoration now requests layout for its anchor character instead of the whole text container.

## Validation

- 55 focused tests passed across display-document, native-input, and editor-integration suites.
- Regression coverage checks localized edits, diagnostic paint removal across hints, atomic fallback for layout changes, and wrapped-line scroll preservation without whole-container layout.
- An isolated optimized-build benchmark with 10,000 lines, 100 hints, and 200 three-character paint updates used about 25% less time. This measures display-document updates, not end-to-end typing latency.
- SwiftFormat and `git diff --check` passed.


mrmans0n:
@codex review

PR #1234, opened 2026-09-15
build: adopt the Swift 6 language mode

mrmans0n:
Adopts the Swift 6 language mode for both the app and the test bundle.

`SWIFT_STRICT_CONCURRENCY` was already `complete`, so the data-race rules here
are not new — they were reported as warnings and are now errors. The bulk of
this branch is clearing those: **615 warnings (293 app + 322 test) to zero**,
then flipping `SWIFT_VERSION` to `6.0`.

## What actually changed

Most of the diff is isolation made explicit rather than behaviour changed.
A few clusters did most of the work:

- **`DiffDisplayRow` gains `Sendable`** — one word cleared 38 warnings across
  ten files. Its self-recursive `collapsedRows` property was blocking implicit
  conformance inference, which cascaded to `DiffDisplayGroup` and
  `DiffDisplayModel`. The conformances are checked, not `@unchecked`.
- **`GitService` and its four loader protocols gain `Sendable`** (~35).
- **`NSLock.lock()/unlock()` in async contexts → `withLock`** (22). No lock was
  held across an `await` anywhere, verified before converting.
- **A shared `NotificationCenter.addMainActorObserver` helper** replaces eight
  copies of the same `MainActor.assumeIsolated` shape.
- **`MergePaneScrollBridge`** extracts the ~90% duplicated scroll observer from
  the two merge panes.

## Two real bugs, not just diagnostics

- `MergeActionGutter` called `MainActor.assumeIsolated` **inside a `deinit`**.
  That is a precondition, not a hop: if the last reference were dropped off the
  main actor it would trap. Now `isolated deinit`.
- `EditorBuffer`'s LSP open task referenced `worktreeRoot` through an implicit
  `self` inside a `[weak self]` closure, capturing `self` strongly and making
  the orphan-cleanup branch unreachable while an open was in flight.

## Behaviour notes for review

- `RightPaneState`'s merge refresh is sequential rather than overlapped.
- Twelve render smoke tests asserted `controller.view != nil`, which is vacuous
  on a non-optional. They now assert the hosted view measures a non-zero
  height rather than asserting nothing.
- `GitLabCLIProvider` gated a `mergeRequestDiffRefs` fetch on
  `$0.filePath != nil`, always true for a non-optional `String`. It now mirrors
  the condition that actually decides inline posting, so a comment with a line
  but an empty path no longer triggers a round-trip it cannot use.
- Run-loop pumping in tests was **not** converted to `Task.sleep`: those sites
  drive `NSHostingController` layout from CFRunLoop observers in default mode,
  and one deliberately blocks the main actor to exercise backpressure.

## Escape hatches

The fix policy was to prefer real fixes and allow `@unchecked Sendable` /
`nonisolated(unsafe)` only with a written invariant. The ones worth a careful
look: `ReviewSessionLoader`, `AgentHookSocketServer` and `HarnessDetector`
(both had their mutable state genuinely relocated under a lock, not just
annotated), and `CodeEditorLayoutManager`.

`DiffReviewLoadedSession` was deliberately **not** made `Sendable` — it stores
closures — so the tests that consumed it are main-actor isolated instead.

## Verification

`xcodebuild build-for-testing` is clean: 0 errors, 0 warnings, both targets,
Swift 6 mode.

`AlasTests/RightPaneGGStackTests` has 14 local failures. These reproduce
identically on the merge base with the same test names, so they are
pre-existing and unrelated to this branch; CI is the authority.


mrmans0n:
@codex review

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #1238, opened 2026-09-15
perf(editor): Update editor line offsets incrementally for source edits

mrmans0n:
## Summary

- Update cached source line starts from validated incremental edits instead of scanning the whole document after ordinary typing.
- Scan only replacement UTF-16 text, splice newline offsets, and shift subsequent starts. Keep full scans for reloads and projection fallback paths.
- Publish matching line offsets before display text-storage observers run.

## Verification

- 62 affected editor projection, input, and integration tests passed, with no failures or skips.
- Added newline boundary, deletion, emoji, CRLF, hinted fallback, synchronous observer, and undo/redo coverage. The new boundary test was verified failing before implementation.
- Focused SwiftFormat lint, xcodegen, and git diff --check passed.
- Independent local review found no critical or important issues.
- Optimized standalone benchmark of the actual index helpers: 1,000 alternating single-character insert/delete updates in a 230 KB, 10,000-line document took 0.63–0.68 seconds with full scans and 0.007–0.008 seconds incrementally across start, middle, and end positions. This measures index work only, not end-to-end typing latency.
- Full build/test coverage is left to CI.


mrmans0n:
CI investigation: the test host crashes in NavigationFeatureTests.definitionPickerUsesSharedDirtySnippetsAndDropsSupersededLoads(cancel:). The same failure is present on main at 423af95c (run 34951905916). An isolated local run reproduces SIGTRAP in the nested map closure of the fake transport callback, with a Swift executor-isolation check. A test-only explicit Sendable callback fix is undergoing focused verification; the line-index tests passed in CI.

mrmans0n:
@codex review Fixed the pre-existing Swift 6 test callback isolation crash with an explicit Sendable closure. The isolated test reproduced SIGTRAP before the change; all 10 NavigationFeatureTests now pass. The original 62 focused display tests and the CI display suites passed. Please review the new head.

PR #1241, opened 2026-09-15
perf(editor): scope editor highlight invalidation to owned display ranges

mrmans0n:
## Summary

- Replace blanket display-wide temporary-attribute clearing with a pre-projection notification so each highlight owner clears only its recorded native ranges before positions change.
- Track semantic foreground, find background, reveal underline, and Cmd-hover underline ranges; reapply only current results after projection updates.
- Preserve unrelated temporary backgrounds and restore backgrounds underneath expired find matches during typing.
- Keep conservative clearing for legacy unprojected semantic/find storage.

## Verification

- 88 focused projection/input/integration, semantic, find, hover, and reveal tests passed with no failures or skips.
- After rebasing onto merged #1239, 98 focused tests passed, including inlay layout and scheduling coverage.
- Verified the background restoration regression failing before implementation.
- Independent review found an unmigrated Cmd-hover owner; added a regression that reproduced the trailing underline after an insertion, fixed it, and verified it passes. Re-review found no further important issues.
- The regression asserts clearing is limited to the owned match range, with no whole-document temporary-attribute removal.
- SwiftFormat lint, xcodegen, and git diff --check passed. Full build/test coverage is left to CI.

This is the separate highlight-invalidation follow-up to merged #1239. The diff contains only highlight ownership changes and their regressions.


mrmans0n:
@codex review Rebased onto merged #1239, preserving both sets of tests. All 98 focused highlight, projection, hover, reveal, and inlay tests passed on the new head. Please review the updated commit.

PR #1291, opened 2026-09-17
fix(files): compacted file tree expansion

mrmans0n:
## Summary
- Keep compacted file tree rows open when lazy loading moves the terminal directory deeper in the compacted chain.
- Add a regression test for preserving expansion when an earlier chain path is open.

## Local checks
- `xcodebuild -quiet -project Alas.xcodeproj -scheme Alas -destination 'platform=macOS,arch=arm64' -derivedDataPath /tmp/alas-files-tab-pr-tests -only-testing AlasTests/FilesTabCompactChainTests test` fails locally before compiling tests because Xcode reports the generated `.build/ghostty/GhosttyKit.xcframework` as missing, despite the artifact and metadata existing after `scripts/build-ghostty.sh` restores it.


mrmans0n:
@codex review

mrmans0n:
@codex review

PR #1296, opened 2026-09-17
feat(config): repo-local .alas/ configuration: shared icon, MCP servers, and default agent

mrmans0n:
## Summary

Adds a team-shared, repo-local configuration layer read from `.alas/` in the repo (design: `docs/plans/2026-09-17-repo-local-config-design.md`, plan: `docs/plans/2026-09-17-repo-local-config-implementation.md`):

- **`.alas/config.json`** (versioned JSON, read-only in v1) can define `mcpServers`, `defaultAgent`, and an optional `icon` (path relative to `.alas/`). A `.alas/icon.png`/`jpg`/`gif`/`webp` is picked up with no config at all.
- **Layering**: global defaults → repo config → per-user project settings. Repo config is a *default*; anything set per project in the app wins, and nothing is written back into the repo.
- **MCP servers merge by name.** A repo-defined server never attaches until the user approves it once (trust keyed by a config hash, so edits re-prompt and identical configs stay approved across branches). App-level servers shadow repo servers by name; repo servers can be disabled per project from the MCP status control. Decisions are per-user and stored in `projects.json`.
- **Default agent**: repo `defaultAgent` applies when the project has no explicit override, and only when the id names an installed, enabled agent. Project override > repo default > global.
- **Icon**: an explicit app icon always wins; a repo icon is staged through the existing content-addressed icon pipeline and the user's color/background preferences carry over. Cache is keyed by the icon file's (path, mtime, size) so re-renders don't re-read or re-hash.
- Remote (SSH-host) projects and frozen checkout snapshots ignore repo config in v1; unknown config keys are ignored and malformed files are treated as absent and logged.

## Notes for reviewers

- All new logic lives in a pure resolution layer (`RepoConfig`, `RepoConfigStore`, `RepoIconResolver`, `RepoMCPResolver`, `RepoMCPTrust`) with Swift Testing suites; UI wiring is thin.
- Repo MCP servers reuse the app's `ProjectMCPTransport` Codable shape (documented in the README), so the existing editor/model is shared.
- Deliberately deferred: a friendlier hand-written server shape (no `id` on env/header entries, `type` alias for `kind`), file watchers, staged-copy eviction, bypass-permissions from repo config (never allowed), and SSH-host support.

PR #1304, opened 2026-09-17
fix(config): confine repo config reads to bounded regular files, use every discovered icon

mrmans0n:
Follow-up to #1296, addressing the two review threads left unresolved when it merged.

## Confine repo config reads to bounded regular files

`.alas/config.json` is now resolved through any symlink components and only read when the canonical target is a regular file inside the repo's `.alas/` directory, with that directory itself required to stay inside the canonical checkout. A committed symlink to a character device such as `/dev/zero` would previously be read on the main actor and block indefinitely on render or attach paths, ahead of the MCP trust prompt. A symlink that stays inside `.alas/` keeps working — it is a normal way to organize repo config files.

The same confinement helper is now shared with the icon resolver, replacing its hand-rolled variant, so config and icon candidates get identical treatment (regular file, contained, canonical path reported back for stat and read).

## Use every discovered icon candidate

`discoveredIconURL` becomes `discoveredIconCandidates` and returns every existing conventional icon in extension order (`png`, `jpg`, `jpeg`, `gif`, `webp`), not just the first. A corrupt or oversized `.alas/icon.png` no longer masks a valid `.alas/icon.jpg`: the resolver stages candidates in order and takes the first usable one.

## Verification

- New tests: symlinked config outside the checkout refused, symlinked config inside `.alas/` followed, `/dev/zero` target refused, every discovered candidate returned, broken `.alas/icon.png` falls through to `.alas/icon.jpg`.
- `RepoConfigStoreTests` + `RepoIconResolverTests` + `AppStateRepoIconTests` + `RepoIconDisplayCacheTests`: 67 tests green locally; `RepoMCPExternalSyncTests`: green.


mrmans0n:
@codex review

PR #1328, opened 2026-09-18
feat(report): Improve run report output viewer

mrmans0n:
## Summary

- Render completed run output in the native read-only AppKit text view.
- Give reports a compact run summary header and keep output copy support.
- Enable native find in the shared read-only viewer and cover report output presentation.

## Testing

Focused RunReportTabTests via xcodebuild.


mrmans0n:
Restored the persisted working directory and endpoint in the compact report header. Both values use selectable text so they remain available for inspection and copy.

PR #1329, opened 2026-09-18
feat(remote-web): redesign the mobile PWA to match the Claude Design canvas

mrmans0n:
## Summary

Restyles the whole remote-web mobile PWA against the "Alas Mobile PWA" design canvas instead of the ad-hoc bordered-card look it had before.

- `style.css`: flat slate tokens (oklch), one teal accent, cards a shade darker than the page — replacing the mixed gradient/border palette.
- Repos list: sticky color-tiled repo headers over worktree cards that show branch, running-agent badge, status dot, diff bar, and relative time, matching the mock. A worktree with a single session opens it directly; one with several nests each session as its own row, with closed sessions under a disclosure.
- Filter chips: "Active" becomes "Running", now based on whether a session is actually mid-turn (streaming/awaiting) rather than just open — added `worktreeIsRunning`/`worktreeStatus`/`worktreeRecencyMs` to `repo-filter.js` with a new pure-function test suite.
- Session screen: nav bar, segmented Chat/Changes/Files tabs, and transcript cards (tool rows, "Thinking…" pips, tool status) restyled to the mock's rounded pill/card language.
- Composer: single rounded field (attach · text · gear) with a separate square send/stop button, matching the mock instead of the old flat inline row.
- Inline SVG icon set replacing mixed emoji/glyphs; manifest theme colors and service worker cache versions bumped to match.

Session titles now show only in the session's own header and in multi-session sub-rows — the design mock treats a single-session worktree card's identity as its branch, not a custom title.

## Test plan

- `bash scripts/tests/remote-web-changes/run.sh`
- `bash scripts/tests/remote-web-file-browser/run.sh`
- `bash scripts/tests/remote-web-repo-filter/run.sh` (new)
- `bash scripts/tests/remote-web-session-ordering/run.sh`
- `bash scripts/tests/remote-web-worktree-creation/run.sh`

No Swift files changed; no local xcodebuild run. CI covers Swift builds/tests.


mrmans0n:
@codex review

mrmans0n:
Fixed in a93d8243: the Repos list now re-requests `listSessions` on a 15s interval while it's visible, mirroring the existing relative-time refresh pattern. `sessionList` is otherwise a pure pull snapshot with no server-side push on streaming-state change, so polling client-side is the smallest complete fix rather than adding a broadcast mechanism to the gateway.

mrmans0n:
@codex review

mrmans0n:
@codex review

PR #1348, opened 2026-09-20
fix(acp): align to the header when a card folds into a collapsed bundle

mrmans0n:
## Symptom

On a live turn, when a finished tool call folded into a collapsed "Ran N tools" bundle, the transcript jumped up to wherever that run of tools began. Scrolling back down did not stick: the next fold jumped it up again. Reported as the tail-anchoring "not taking into account that the tools are collapsed".

## Root cause

`resolveStaleRowId` reported **head growth** for a card that had just become its bundle's last member, on the reasoning that the card's bottom edge now coincided with the bundle's. That is only true when the bundle row *contains* the card's rendered height.

A collapsed header is one line regardless of member count, so the card's height simply left the document. `restoreViaStaleResolution` then did bottom-relative restoration from the **old** card's height, clamped to the **new** header's height, which pins the viewport to within one line of the header's bottom edge. For a long run of tools that is far above the tail.

It recurred on every fold because a scroll gesture in flight suppresses the tail re-pin for a short window, and that is exactly when the reader is trying to get back to the composer.

## Fix

A member's own stale id can only ever arise from a **collapsed** bundle, since an expanded bundle tiles each member under its own row id (#1340). So the plain-id branch of the resolver now never assumes head growth. Restoration falls to the top-relative, capped path every row that shrinks in place already gets: the viewport aligns to the header's top, capped at the header's height, and the first thing on screen is what followed the bundle.

The re-keyed-group branch is unchanged. There the replacement genuinely is another bundle row, which is the case bottom-relative math was written for.

## Tests

- Flipped `staleResolutionForBundledPlainIdAsLastMember` from asserting head growth to asserting none, with the reasoning rewritten. Watched it fail on the old code first.
- Added `removalIntoCollapsedBundleAlignsTopRelative` in the reconciler apply suite: a 400pt card scrolled 300pt into, absorbed into a 20pt header that does not grow, ends with the viewport at the header's bottom edge. The existing neighbour test models a *growing* bundle row, which no longer occurs for a plain member post-#1340, but still pins the reconciler's contract for the head-growth flag and is left as is.

Local run, all green:

```
xcodebuild ... -only-testing AlasTests/ACPTranscriptVisibleRowLookupGroupTests \
  -only-testing AlasTests/ACPTranscriptScrollerReconcilerApplyTests \
  -only-testing AlasTests/ACPTranscriptScrollerReconcilerDiffTests \
  -only-testing AlasTests/ACPToolCallGroupingTests \
  -only-testing AlasTests/ACPTranscriptScrollerPolicyTests \
  -only-testing AlasTests/ACPTranscriptScrollerScrollBackTests \
  -only-testing AlasTests/ACPTranscriptScrollerConvergenceTests test
# Test run with 103 tests in 7 suites passed
```


mrmans0n:
@codex review

PR #1357, opened 2026-09-21
fix(acp): stop tool-call bursts from trimming short context out of the transcript window

mrmans0n:
## Problem

A live turn dominated by a long run of tool calls could silently scroll the
transcript so it looked like only "Ran N tools" and a couple of short lines
were left — hiding the user's own prompt above the fold — even though the
whole conversation would easily fit on one screen. Scrolling back to the
top would briefly reveal everything, but it snapped back within about a
second, repeatedly, for as long as tools kept running.

## Root cause

`ACPSession.advanceRenderWindowIfFollowingTail` re-runs on every appended
transcript message while the session follows the tail, and sized the
render window (`ACPTranscript.tailWindow`, 30) by raw message count. Each
individual tool call is its own transcript message, but a whole run of them
collapses into a single "Ran N tools" row in the UI. A turn that ran a few
dozen tools could by itself exhaust the entire 30-message budget, pushing
`visibleHead` far enough forward to trim the user's prompt (and anything
else before the run) out of `transcript.messages` — even though the
rendered result was just a handful of short rows. Because this reran on
every appended message, the trim recurred continuously through the run
instead of settling once, which is why the position kept snapping back.

## Fix

`ACPTranscript.tailWindowHead` now weights a maximal contiguous run of
tool-call messages as a single budget unit instead of one unit per call —
matching how it actually renders. Both `resetWindowToTail` (used after
hydration) and the live tail-follow advance now go through this instead of
the flat `messages.count - tailWindow` subtraction.

## Testing

- `AlasTests/ACPTranscriptWindowTests` — new pure-model coverage for
  `tailWindowHead`: a tool-call run that dominates the window no longer
  trims short preceding context, weighting is still bounded correctly once
  enough real content precedes the run, and it never does *worse* than the
  old per-message baseline.
- `AlasTests/ACPSessionTests` — new end-to-end regression test through the
  real `apply()` update path: a 43-tool-call burst no longer trims the
  user's prompt out of the live render window.
- Full local build (`xcodebuild ... build`) succeeded.
- Focused suites (`ACPTranscriptWindowTests`, `ACPSessionTests`,
  `ACPSessionManagerHydrationTests`, `ACPSessionManagerLeaseTests`,
  `ACPSessionManagerAttachRestoreTests`,
  `ACPTranscriptMarkdownCacheEvictionTests`) — 334 tests passed.


mrmans0n:
Fixed in 3a55547f: `tailWindowHead` now caps how much of one contiguous tool-call run it absorbs for free at `maxVisibleRows` raw messages per budget unit, so a run can never soak up more than `tailWindow * maxVisibleRows` messages regardless of the `collapsesFinishedToolCalls` setting. Added a regression test (`tailWindowHeadCapsAPathologicallyLongRun`) confirming a run well past that cap still yields a nonzero head.

mrmans0n:
Fixed in a68735c8: removed the agent attribution from the test comment.

Issue #1162, opened 2026-09-10
Add scheduled run scripts and agent launches


mrmans0n:
## Problem

Run scripts and agent launches are manual actions. There is no way to tell Alas "run this script every morning", "run the test suite when I start the day", or "open a fresh worktree with an agent when CI on main fails". Users who want this today write their own cron jobs that call the `alas` CLI from outside the app, with no first-class state, notifications, or history.

## Proposed experience

- Schedule a run script at an interval or time of day, scoped globally, per project, or per worktree.
- Optional post-run composition with existing machinery: a scheduled run may create a worktree and auto-launch an agent on it, reusing the worktree-create script and agent auto-launch pipelines.
- Schedules survive relaunch and are honest about sleep: define a missed-run policy per schedule (skip missed occurrences or run the latest one) and show when the machine was asleep, instead of silently firing a burst of catch-up runs.
- Run outcomes flow into the existing completion notifications and any future event history; a scheduled run is indistinguishable from a manual one once started.
- A schedules list showing each schedule, its target (local or SSH host), last outcome, and next fire time, with enable/disable and a run-now action.
- Schedules can be paused per project or globally, for example while on vacation.

Scheduling is about *when* work starts; it must reuse the existing execution paths for *how* work runs, including repository/global script discovery and SSH execution.

## Scope and delivery

1. Deliver interval and time-of-day schedules for run scripts, with persistence, missed-run policy, and the schedules list.
2. Add composition: schedule → new worktree → agent auto-launch, as an option on a schedule.
3. Integrate outcomes with notifications and completion records.

External triggers (webhooks, CI failures, issue assignment), natural-language schedule entry, and scheduling ACP chat sessions as opposed to terminal agents are follow-ups, not prerequisites. Do not build a general-purpose cron daemon or run work while the app is quit.

## Acceptance criteria

- [ ] A user can create, edit, enable/disable, and delete a schedule, and see last outcome and next fire time.
- [ ] Schedules survive relaunch; missed occurrences follow the configured policy and never silently stack.
- [ ] A scheduled run executes through the same path as a manual run and reports the same outcomes, including failures and captured output.
- [ ] Worktree + agent composition reuses the existing worktree-create script and agent auto-launch resolution, and the resulting session is attributed to the new worktree.
- [ ] Local and SSH targets run on the correct host with explicit attribution.
- [ ] No scheduled work executes while the app is not running; the UI states this plainly.
[...]

PR #1358, opened 2026-09-21
feat(schedules): remember each firing and link its run report

mrmans0n:
Follow-ups to #1162 / #1353.

## 1. Schedule firing history

Schedules kept only their last outcome, so "did the nightly actually run on Tuesday?" had no answer. Each schedule now keeps its last 20 firings, shown in a collapsed **History** disclosure on the card.

Each entry records when it fired, how long it took, whether it was a Run Now, and what it produced.

- **No duplicated transcripts.** A firing stores the run's coordinates and the row opens the existing run report, so `RunHistoryStore` stays the single owner of output and of its retention.
- **Deliberate non-runs are recorded**: a batch of missed occurrences dropped by the `skip` policy, and an occurrence refused because the previous run was still going. A quiet night reads as a decision rather than a hole.
- **No dangling links.** A launch that never produced a command links nothing, because its record was rolled back. A run whose report was later purged is still named but is not offered as something to open.
- Opening the tab primes the report ids for worktrees the history points at, so links still work after a relaunch without first visiting each worktree's Run tab.

The runner returns `RunScheduleRunReport` instead of a bare outcome to carry run ids back to the scheduler. `RunScheduleState` decodes by hand because synthesized `Decodable` ignores property defaults, and a file written before `firings` existed must not lose `nextFireAt`.

## 2. The local test hang, diagnosed and fixed

The second follow-up was "the composition tests wedge locally." They do, and the cause is not the worktrees.

`AppState.fileActionErrorHandler` defaults to `NSAlert.runModal()`. In a test host nobody can dismiss that alert, so the main thread blocks for the rest of the run: the suite stops dead, reports no failure, and produces no further output. Sampling the stuck process showed the main thread in `archiveFinalizedRun` → `showFileActionError` → `runModal` → `_doModalLoop`.

The trigger here was a teardown race: the composition fixture put the run-history sqlite inside a temp repo that the test's `defer` deleted while an archive was still writing. sqlite logged `vnode unlinked while in use`, the write failed, and the error became a modal. Because that archive is trailing work of a test that already passed, the alert lands during *whichever test runs next* — which is why it looked positional.

Fixes:

- `showWarningAlert` no-ops and logs under a test host. Most `AlasTests` files construct `AppState` without injecting a handler, so this closes the hang for all of them.
- The composed fixture injects a handler that reports any alert as a test issue.
- The composition tests flush run history before teardown.

Still open, deliberately: 25 other blocking modal sites remain (16 `alert.runModal()` confirmations, 9 file panels). They return values, so suppressing them would silently answer confirmations.

## Verification

Local, this machine:

| Suites | Before | After |
| --- | --- | --- |
| `RunSchedulePlannerTests`, `RunSchedulerTests`, `RunSchedulePresentationTests`, `AppStateRunScheduleTests` | hung indefinitely | 74 tests pass in 2.4s |
| `AppStateCleanupTests`, `AppStateCLIRoutingTests`, `AppStateCreateWorktreeLaunchSurfaceTests`, `AppStateCreateWorktreeSymlinkTests`, `AppStateSpacesTests` | documented as "unrunnable locally" | 131 tests in 35s, nothing hangs |

In that second set, four suites pass outright. `AppStateCleanupTests` has 3 failing ACP-session tests. Those are pre-existing local failures that were previously hidden behind the hang, not regressions from this change — nothing here touches ACP. CI is the judge of them.

New tests cover recording order, the 20-entry bound, manual marking, missed-batch and collision entries, relaunch round-trip, a state written before `firings` existed, one undecodable firing, run linking for failed and composed runs, report priming, and the label formatting.


mrmans0n:
@codex review

mrmans0n:
Fixed in b25ebae8 — good catch, and it was exactly right.

`activateWorktreeCenterTab` only calls `tabs.activate(worktreeId:tabId:)`; nothing in that path touches `selectedWorktreeId`, and there is no separate project selection, so the centre pane kept rendering whatever worktree the user was on. For a composed schedule the run is *never* in the card's worktree, so the link would have looked dead every time.

The action now goes through `AppState.openScheduleFiringRun(_:)`, which calls `selectWorktree(id:)` before `openRunReport`. Background firing still leaves the selection alone; only following a link moves it, since that is an explicit request to go there.

Covered by `openingAFiringRunSelectsThatRunsWorktree`, which runs a composed schedule from the main worktree, asserts the firing did not steal the selection, then asserts that opening the linked run both moves `selectedWorktreeId` to the created worktree and puts a run-report tab there.

Local: `RunSchedulePlannerTests`, `RunSchedulerTests`, `RunSchedulePresentationTests`, `AppStateRunScheduleTests` — 75 tests, all pass.

mrmans0n:
@codex review

mrmans0n:
Fixed in 7f8a7ad8 — confirmed, and the reasoning holds exactly as described.

`finishArchivingWorktree` calls `cleanupWorktreeState(worktreeId:purgeRunHistory: false)`, so the report survives, and priming reloads its id, so `hasRunReport` stayed true. Meanwhile `CenterSelectionState.candidateWorktrees` resolves only `projectsManager.visibleWorktrees(projectId:)`, which filters out the hidden set. Selecting an archived id would have emptied the centre pane and stranded the tab.

I took the "do not offer the link" branch rather than unarchiving. Unarchiving would change the user's sidebar as a side effect of clicking a history row, which is a bigger action than the click asks for. The entry still names the run; it just stops being a link, the same way a purged report already did. That is now the single gate `canOpenScheduleFiringRun`: the report must exist *and* its worktree must be resolvable.

Covered by `anArchivedWorktreesRunIsNamedButNotOffered`, which runs a composed schedule, asserts the link is offered, hides the created worktree via `setWorktreeHidden`, then asserts `hasRunReport` is still true while `canOpenScheduleFiringRun` is false.

Same push also fixes a race I found in `aLongRunningTargetDoesNotStarveTheOtherProjects`: it waited for the second target to reach `.running` and then asserted the first was too. The targets start as independent tasks, so the suite lost that race whenever the second won. It now waits for both.

Local: 76 tests across `RunSchedulePlannerTests`, `RunSchedulerTests`, `RunSchedulePresentationTests`, `AppStateRunScheduleTests` — all pass.

mrmans0n:
@codex review

mrmans0n:
Both fixed in cd00c4da. Both were right.

**Spaces.** Confirmed: `RootView` resolves the centre pane through the active Space's projects, so moving `selectedWorktreeId` alone leaves it empty whenever an all-projects firing lands in a project outside the active Space. Opening a firing's run now goes through `focusGlobalWorktree(id:projectId:)`, which switches to the containing Space and then selects — exactly the existing path you pointed at. The project is resolved by the same visible-worktree lookup `canOpenScheduleFiringRun` uses, so the two can never disagree about which worktree is reachable.

**Unarchive.** Confirmed: `cleanupWorktreeState` sets `durableRunReportIDsByWorktreeID[worktreeID] = []` unconditionally, including on the `purgeRunHistory: false` archive path, and nothing reloads it on unarchive. My `== nil` guard then treated the cleared cache as primed. Empty now counts as unloaded rather than loaded-and-known-empty: every worktree that reaches priming has at least one firing run, so an empty entry is always worth one query.

New tests: `openingAFiringRunSwitchesToItsSpace` puts the project in a non-active Space and asserts the active Space moves along with the selection; `unarchivingAWorktreeRestoresItsHistoryLinks` reproduces the cleared-cache state, asserts the link is refused, primes, and asserts it comes back.

Local: 78 tests across the four schedule suites — all pass.

mrmans0n:
@codex review

mrmans0n:
Fixed in 44f762ae. Right again, and it is a genuinely separate hole from the cache-guard one.

Making empty mean "unloaded" only helps if the task runs. The identity was `historyWorktreeIDs`, which is derived purely from persisted firing references, so archiving and unarchiving left it byte-identical and SwiftUI never restarted the task.

Visibility is now part of the identity. The token is `"<worktreeID>:<0|1>"` per referenced worktree, where the flag comes from `visibleProjectForWorktree(_:)` — the same lookup `canOpenScheduleFiringRun` and `openScheduleFiringRun` use. That is deliberate: the link, the navigation target and the priming trigger now read the same source, so they cannot disagree about which worktrees are reachable. `projectsManager` is observable, so unarchiving re-evaluates the body, the token flips, and the task re-fires.

`anArchivedWorktreesRunIsNamedButNotOffered` now covers the full round trip: link offered, archive, report still present but link refused and `visibleProjectForWorktree` nil, unarchive, signal and link both restored.

Local: 78 tests across the four schedule suites — all pass.

mrmans0n:
@codex review

PR #1384, opened 2026-09-21
Restart legacy ACP brokers missing todo snapshots

mrmans0n:
Follow-up to #1373.\n\nPreserves Cursor todo merge semantics when an app instance adopts a broker started before todo snapshot support existed by restarting that legacy broker before replay.


mrmans0n:
Superseded by a clean follow-up branch based on current main.

